PR #1, opened 2020-03-03
Free symbol is required on WebAssembly runtime

Shikugawa:
I can't run a WebAssembly filter built with this SDK. Because the assembly doesn't include `<free>` symbol that is required on WebAssembly runtime.


PiotrSikora:
It's not required anymore, it was removed in https://github.com/envoyproxy/envoy-wasm/pull/404.

PR #11, opened 2020-05-11
bazel support

Shikugawa:
Introduced bazel support to build proxy_wasm rust implementation.


src/dispatcher.rs
```diff
@@ -15,7 +15,7 @@
 use crate::hostcalls;
 use crate::traits::*;
 use crate::types::*;
-use hashbrown::HashMap;
+use std::collections::HashMap;
```
Shikugawa, line 18: When we use hashbrown here in bazel, we should resolve all of dependencies with hand-written build definitions. It may be complicated. Is there a critical way to overcome it? 
In the same reason, we'd not like to build `examples/hello_world.rs` because of `chrono`.

PiotrSikora:
Merged with fixes in #18. Thanks!

PR #24, opened 2020-08-10
Add standard Rust attributes.

yskopets:
## Summary

* add [#[non_exhaustive]](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) attribute to enums to reserve an option to add more values in a backwards-compatible way
* add `#[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]`


PiotrSikora:
While technically correct, I'm wondering why do you need `#[non_exhaustive]`?

Also, once we move to the new ABI, those values will be broken in non-backwards-compatible way, so I'm not sure if adding those guarantees and attributes now is such a good idea.

yskopets:
It's about reserving a room for changes that otherwise would be backwards incompatible.

The problem comes form `match`  statements.

Without `#[non_exhaustive]`, if a user creates a `match`  statement with all values of `Status`, then adding a new `Status` value will be a breaking change.

You will have to do a major release of the library to stick to semantic versioning.

lizan:
@yskopets I don't think we need non-exaustive for all enums, for some of them it doesn't make sense. If ABI changes, those match statements need to be rewritten or it will break at runtime.

src/types.rs
```diff
@@ -19,7 +19,8 @@ pub type NewStreamContext = fn(context_id: u32, root_context_id: u32) -> Box<dyn
 pub type NewHttpContext = fn(context_id: u32, root_context_id: u32) -> Box<dyn HttpContext>;
 
 #[repr(u32)]
-#[derive(Debug)]
+#[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
+#[non_exhaustive]
```
PiotrSikora, line 23: This is unlikely to change, and we should definitely allow exhaustive `match` on `LogLevel`.

yskopets:
> to why do we need PartialEq, Eq and Hash

Overall, implementing these traits is advised by the [Rust API Guidelines (interoperability)](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits).

Practically, `Eq` and `PartialEq` come in pair and necessary for assertions like

```rust
#[test]
fn test_log_level_eq() {
    assert_ne!(LogLevel::Critical, LogLevel::Debug);
}
```

All 3 `Eq`, `PartialEq`, `Hash` are mostly necessary in the context of user-defined composite types, which are common in unit testing.

src/types.rs
yskopets: Feel free to replace this PR or edit it in place.

PiotrSikora (approved):
Thanks!

FWIW, I still think that `#[non_exhaustive]` shouldn't be necessary long-term, but the gRPC PRs (#100, #101) added new values, so I'm going to merge this to avoid breaking existing plugins.

PR #30, opened 2020-08-14
Add metrics.

### LINKED ISSUE #4: Rust SDK is missing metrics API

PiotrSikora:
Fixes #4.

### DIFF
--- a/src/hostcalls.rs
+++ b/src/hostcalls.rs
@@ -698,11 +698,11 @@ pub fn define_metric(metric_type: MetricType, name: &str) -> Result<u32, Status>
 }
 
 extern "C" {
-    fn proxy_get_metric(metric_id: u32, return_value: *mut i64) -> Status;
+    fn proxy_get_metric(metric_id: u32, return_value: *mut u64) -> Status;
 }
 
-pub fn get_metric(metric_id: u32) -> Result<i64, Status> {
-    let mut return_value: i64 = 0;
+pub fn get_metric(metric_id: u32) -> Result<u64, Status> {
+    let mut return_value: u64 = 0;
     unsafe {
         match proxy_get_metric(metric_id, &mut return_value) {
             Status::Ok => Ok(return_value),
@@ -713,10 +713,10 @@ pub fn get_metric(metric_id: u32) -> Result<i64, Status> {
 }
 
 extern "C" {
-    fn proxy_record_metric(metric_id: u32, value: i64) -> Status;
+    fn proxy_record_metric(metric_id: u32, value: u64) -> Status;
 }
 
-pub fn record_metric(metric_id: u32, value: i64) -> Result<(), Status> {
+pub fn record_metric(metric_id: u32, value: u64) -> Result<(), Status> {
     unsafe {
         match proxy_record_metric(metric_id, value) {
             Status::Ok => Ok(()),
--- a/src/hostcalls.rs
+++ b/src/hostcalls.rs
@@ -678,6 +678,66 @@ pub fn done() -> Result<(), Status> {
     }
 }
 
+extern "C" {
+    fn proxy_define_metric(
+        metric_type: MetricType,
+        name_data: *const u8,
+        name_size: usize,
+        return_id: *mut u32,
+    ) -> Status;
+}
+
+pub fn define_metric(metric_type: MetricType, name: &str) -> Result<u32, Status> {
+    let mut return_id: u32 = 0;
+    unsafe {
+        match proxy_define_metric(metric_type, name.as_ptr(), name.len(), &mut return_id) {
+            Status::Ok => Ok(return_id),
+            status => panic!("unexpected status: {}", status as u32),
+        }
+    }
+}
+
+extern "C" {
+    fn proxy_get_metric(metric_id: u32, return_value: *mut i64) -> Status;
+}
+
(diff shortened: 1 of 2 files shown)


src/hostcalls.rs
```diff
@@ -678,6 +678,66 @@ pub fn done() -> Result<(), Status> {
...
+    unsafe {
+        match proxy_define_metric(metric_type, name.as_ptr(), name.len(), &mut return_id) {
+            Status::Ok => Ok(return_id),
+            status => panic!("unexpected status: {}", status as u32),
+        }
+    }
+}
+
+extern "C" {
+    fn proxy_get_metric(metric_id: u32, return_value: *mut i64) -> Status;
```
yskopets, line 701: This operation is defined in `proxy-wasm-cpp-host` using `u64` (https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/e73c498e64c654c06a68ea54243e0cc9e2299464/include/proxy-wasm/exports.h#L84)

Why `i64` here ?

src/hostcalls.rs
```diff
@@ -678,6 +678,66 @@ pub fn done() -> Result<(), Status> {
...
+        match proxy_get_metric(metric_id, &mut return_value) {
+            Status::Ok => Ok(return_value),
+            Status::BadArgument => Err(Status::BadArgument),
+            status => panic!("unexpected status: {}", status as u32),
+        }
+    }
+}
+
+extern "C" {
+    fn proxy_record_metric(metric_id: u32, value: i64) -> Status;
```
yskopets, line 716: This operation is defined in `proxy-wasm-cpp-host` using `u64` (https://github.com/proxy-wasm/proxy-wasm-cpp-host/blob/e73c498e64c654c06a68ea54243e0cc9e2299464/include/proxy-wasm/exports.h#L83)

Why `i64` here ?

src/hostcalls.rs
PiotrSikora: Mistake. Thanks for catching it!

src/hostcalls.rs
PiotrSikora: Mistake. Thanks for catching it!

src/hostcalls.rs
```diff
@@ -678,6 +678,66 @@ pub fn done() -> Result<(), Status> {
...
+    fn proxy_get_metric(metric_id: u32, return_value: *mut u64) -> Status;
+}
+
+pub fn get_metric(metric_id: u32) -> Result<u64, Status> {
+    let mut return_value: u64 = 0;
+    unsafe {
+        match proxy_get_metric(metric_id, &mut return_value) {
+            Status::Ok => Ok(return_value),
+            Status::BadArgument => Err(Status::BadArgument),
+            status => panic!("unexpected status: {}", status as u32),
```
yskopets, line 710: I'm looking into implementation of this callback on Envoy side (https://github.com/envoyproxy/envoy-wasm/blob/master/source/extensions/common/wasm/context.cc#L1250-L1268).

It can return one of `Ok / BadArgument / NotFound`.

Why only `BadArgument` is handled here but not `NotFound` ?

src/hostcalls.rs
```diff
@@ -678,6 +678,66 @@ pub fn done() -> Result<(), Status> {
...
+        return_id: *mut u32,
+    ) -> Status;
+}
+
+pub fn define_metric(metric_type: MetricType, name: &str) -> Result<u32, Status> {
+    let mut return_id: u32 = 0;
+    unsafe {
+        match proxy_define_metric(metric_type, name.as_ptr(), name.len(), &mut return_id) {
+            Status::Ok => Ok(return_id),
+            status => panic!("unexpected status: {}", status as u32),
```
yskopets, line 695: Why `BadArgument ` is not handled here similarly to `get_metric ` ?

src/hostcalls.rs
```diff
@@ -678,6 +678,66 @@ pub fn done() -> Result<(), Status> {
...
+
+extern "C" {
+    fn proxy_record_metric(metric_id: u32, value: u64) -> Status;
+}
+
+pub fn record_metric(metric_id: u32, value: u64) -> Result<(), Status> {
+    unsafe {
+        match proxy_record_metric(metric_id, value) {
+            Status::Ok => Ok(()),
+            status => panic!("unexpected status: {}", status as u32),
```
yskopets, line 723: Why `NotFound ` is not handled here similarly to `get_metric ` ?

src/hostcalls.rs
```diff
@@ -678,6 +678,66 @@ pub fn done() -> Result<(), Status> {
...
+
+extern "C" {
+    fn proxy_increment_metric(metric_id: u32, offset: i64) -> Status;
+}
+
+pub fn increment_metric(metric_id: u32, offset: i64) -> Result<(), Status> {
+    unsafe {
+        match proxy_increment_metric(metric_id, offset) {
+            Status::Ok => Ok(()),
+            status => panic!("unexpected status: {}", status as u32),
```
yskopets, line 736: Why `BadArgument / NotFound ` are not handled here similarly to `get_metric ` ?

src/hostcalls.rs
PiotrSikora: I was playing around with a higher-level object wrappers to metrics, where `metric_id` is never exposed to the consumer, where `NotFound` couldn't happen, and I forgot to add it to the `hostcalls`.

Thanks for catching it!

PR #32, opened 2020-08-14
Fix HTTP body example.

PiotrSikora:

### DIFF
--- a/examples/http_body.rs
+++ b/examples/http_body.rs
@@ -18,19 +18,10 @@ use proxy_wasm::types::*;
 #[no_mangle]
 pub fn _start() {
     proxy_wasm::set_log_level(LogLevel::Trace);
-    proxy_wasm::set_http_context(|_, _| -> Box<dyn HttpContext> { Box::new(HttpBody::new()) });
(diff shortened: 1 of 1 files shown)


examples/http_body.rs
```diff
@@ -41,13 +32,10 @@ impl HttpContext for HttpBody {
...
         // can no longer modify the response headers.
         self.set_http_response_header("content-length", None);
-        // Don't continue to the next callout in the chain because we might
-        // modify the body.
-        Action::Pause
+        Action::Continue
     }
 
     fn on_http_response_body(&mut self, body_size: usize, end_of_stream: bool) -> Action {
-        self.total_body_size += body_size;
```
PiotrSikora, line 50: Hi @gbrail, I'm not sure how I've missed this change while reviewing your PR. I think I just assumed that you've applied suggested changes and I merged it without thinking... In any case, `on_http_response_body` reports total buffered body size, not only the most recent chunk (that was a regression in Envoy-Wasm, see: https://github.com/envoyproxy/envoy-wasm/pull/608).

gbrail: Yes -- before the callback was getting only the size of the current chunk -- with this fix the code is much simpler. Thanks.

PR #34, opened 2020-08-21
Allow RootContext to create child contexts for streams.

### LINKED ISSUE #6: Unlike in other SDKs, there is no link between `RootContext` and `HttpContext` / `StreamContext`
## Summary

* Due to a missing link between `RootContext` and `HttpContext`, it's not possible to to apply to `HttpContext` configuration that was set in the `RootContext` during `proxy_on_configure`
* Why is `HttpContext` gets created by a standalone factory function rather than by `RootContext` ?

dgn:
To fix #6 

Without this, it's impossible (or just incredibly hard, I tried for hours) to read filter config from within an `HttpContext` or `StreamContext`. It requires users to always supply a `RootContext`, which adds a bit of boilerplate, but IMO is a reasonable way to connect `RootContext` with its children in Rust, as there's no easy way to manage global state.

I'm still very new to Rust so this might be the wrong way to do it. Let's discuss.

### DIFF
--- a/examples/http_auth_random.rs
+++ b/examples/http_auth_random.rs
@@ -20,21 +20,7 @@ use std::time::Duration;
 #[no_mangle]
 pub fn _start() {
     proxy_wasm::set_log_level(LogLevel::Trace);
-    proxy_wasm::set_root_context(|_| -> Box<dyn RootContext> { Box::new(HttpAuthRandomRoot) });
-}
-
-struct HttpAuthRandomRoot;
-
-impl Context for HttpAuthRandomRoot {}
-
-impl RootContext for HttpAuthRandomRoot {
-    fn get_type(&self) -> ContextType {
-        ContextType::HttpContext
-    }
-
-    fn create_http_context(&self, _context_id: u32) -> Box<dyn HttpContext> {
-        Box::new(HttpAuthRandom)
-    }
+    proxy_wasm::set_http_context(|_, _| -> Box<dyn HttpContext> { Box::new(HttpAuthRandom) });
 }
 
 struct HttpAuthRandom;
--- a/examples/http_auth_random.rs
+++ b/examples/http_auth_random.rs
@@ -20,10 +20,23 @@ use std::time::Duration;
 #[no_mangle]
 pub fn _start() {
     proxy_wasm::set_log_level(LogLevel::Trace);
-    proxy_wasm::set_http_context(|_, _| -> Box<dyn HttpContext> { Box::new(HttpAuthRandom) });
+    proxy_wasm::set_root_context(|_| -> Box<dyn RootContext> { Box::new(HttpAuthRandomRoot) });
 }
 
 struct HttpAuthRandom;
+struct HttpAuthRandomRoot;
+
+impl RootContext for HttpAuthRandomRoot {
+    fn get_type(&self) -> ContextType {
+        ContextType::HttpContext
+    }
+
+    fn create_http_context(&self, _root_context_id: u32, _context_id: u32) -> Box<dyn HttpContext> {
+        Box::new(HttpAuthRandom)
+    }
+}
+
+impl Context for HttpAuthRandomRoot {}
 
 impl HttpContext for HttpAuthRandom {
     fn on_http_request_headers(&mut self, _: usize) -> Action {
--- a/examples/http_auth_random.rs
+++ b/examples/http_auth_random.rs
@@ -25,24 +25,6 @@ pub fn _start() {
 
 struct HttpAuthRandom;
 
-impl Context for HttpAuthRandom {
-    fn on_http_call_response(&mut self, _: u32, _: usize, body_size: usize, _: usize) {
-        if let Some(body) = self.get_http_call_response_body(0, body_size) {
-            if !body.is_empty() && body[0] % 2 == 0 {
-                trace!("Access granted.");
-                self.resume_http_request();
-                return;
-            }
-        }
-        trace!("Access forbidden.");
-        self.send_http_response(
-            403,
-            vec![("Powered-By", "proxy-wasm")],
-            Some(b"Access forbidden.\n"),
-        );
-    }
-}
-
 impl HttpContext for HttpAuthRandom {
     fn on_http_request_headers(&mut self, _: usize) -> Action {
         self.dispatch_http_call(
@@ -63,5 +45,23 @@ impl HttpContext for HttpAuthRandom {
     fn on_http_response_headers(&mut self, _: usize) -> Action {
         self.set_http_response_header("Powered-By", Some("proxy-wasm"));
         Action::Continue
+    }
+}
+
+impl Context for HttpAuthRandom {
+    fn on_http_call_response(&mut self, _: u32, _: usize, body_size: usize, _: usize) {
+        if let Some(body) = self.get_http_call_response_body(0, body_size) {
+            if !body.is_empty() && body[0] % 2 == 0 {
+                trace!("Access granted.");
+                self.resume_http_request();
+                return;
+            }
+        }
+        trace!("Access forbidden.");
+        self.send_http_response(
+            403,
+            vec![("Powered-By", "proxy-wasm")],
+            Some(b"Access forbidden.\n"),
+        );
     }
 }
--- a/examples/http_auth_random.rs
+++ b/examples/http_auth_random.rs
@@ -23,7 +23,6 @@ pub fn _start() {
     proxy_wasm::set_root_context(|_| -> Box<dyn RootContext> { Box::new(HttpAuthRandomRoot) });
 }
 
-struct HttpAuthRandom;
 struct HttpAuthRandomRoot;
 
 impl Context for HttpAuthRandomRoot {}
@@ -37,6 +36,8 @@ impl RootContext for HttpAuthRandomRoot {
         Box::new(HttpAuthRandom)
     }
 }
+
+struct HttpAuthRandom;
 
 impl Context for HttpAuthRandom {
     fn on_http_call_response(&mut self, _: u32, _: usize, body_size: usize, _: usize) {
--- a/examples/http_auth_random.rs
+++ b/examples/http_auth_random.rs
@@ -26,17 +26,35 @@ pub fn _start() {
 struct HttpAuthRandom;
 struct HttpAuthRandomRoot;
 
+impl Context for HttpAuthRandomRoot {}
+
 impl RootContext for HttpAuthRandomRoot {
     fn get_type(&self) -> ContextType {
         ContextType::HttpContext
     }
 
-    fn create_http_context(&self, _root_context_id: u32, _context_id: u32) -> Box<dyn HttpContext> {
+    fn create_http_context(&self, _context_id: u32) -> Box<dyn HttpContext> {
         Box::new(HttpAuthRandom)
     }
 }
 
-impl Context for HttpAuthRandomRoot {}
+impl Context for HttpAuthRandom {
+    fn on_http_call_response(&mut self, _: u32, _: usize, body_size: usize, _: usize) {
+        if let Some(body) = self.get_http_call_response_body(0, body_size) {
+            if !body.is_empty() && body[0] % 2 == 0 {
+                trace!("Access granted.");
+                self.resume_http_request();
+                return;
+            }
+        }
+        trace!("Access forbidden.");
+        self.send_http_response(
+            403,
+            vec![("Powered-By", "proxy-wasm")],
+            Some(b"Access forbidden.\n"),
+        );
+    }
+}
 
 impl HttpContext for HttpAuthRandom {
     fn on_http_request_headers(&mut self, _: usize) -> Action {
@@ -58,23 +76,5 @@ impl HttpContext for HttpAuthRandom {
     fn on_http_response_headers(&mut self, _: usize) -> Action {
         self.set_http_response_header("Powered-By", Some("proxy-wasm"));
         Action::Continue
-    }
-}
-
-impl Context for HttpAuthRandom {
-    fn on_http_call_response(&mut self, _: u32, _: usize, body_size: usize, _: usize) {
-        if let Some(body) = self.get_http_call_response_body(0, body_size) {
-            if !body.is_empty() && body[0] % 2 == 0 {
-                trace!("Access granted.");
-                self.resume_http_request();
-                return;
-            }
-        }
-        trace!("Access forbidden.");
-        self.send_http_response(
-            403,
-            vec![("Powered-By", "proxy-wasm")],
-            Some(b"Access forbidden.\n"),
-        );
     }
 }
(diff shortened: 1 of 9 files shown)


PiotrSikora (changes requested):

src/dispatcher.rs
```diff
@@ -96,12 +76,9 @@ impl Dispatcher {
...
     fn create_stream_context(&self, context_id: u32, root_context_id: u32) {
-        if !self.roots.borrow().contains_key(&root_context_id) {
-            panic!("invalid root_context_id")
-        }
-        let new_context = match self.new_stream.get() {
-            Some(f) => f(context_id, root_context_id),
-            None => panic!("missing constructor"),
+        let new_context = match self.roots.borrow().get(&root_context_id) {
+            Some(root_context) => root_context.create_stream_context(context_id, root_context_id),
+            None => panic!("invalid root_context_id"),
```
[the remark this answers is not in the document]
dgn: I moved all context-mgmt/creation into `on_create_context` now

src/dispatcher.rs
```diff
@@ -114,12 +91,9 @@ impl Dispatcher {
...
     fn create_http_context(&self, context_id: u32, root_context_id: u32) {
-        if !self.roots.borrow().contains_key(&root_context_id) {
-            panic!("invalid root_context_id")
-        }
-        let new_context = match self.new_http_stream.get() {
-            Some(f) => f(context_id, root_context_id),
-            None => panic!("missing constructor"),
+        let new_context = match self.roots.borrow().get(&root_context_id) {
+            Some(root_context) => root_context.create_http_context(context_id, root_context_id),
+            None => panic!("invalid root_context_id"),
```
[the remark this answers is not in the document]
dgn: I moved all context-mgmt/creation into `on_create_context` now

src/dispatcher.rs
```diff
@@ -144,11 +118,15 @@ impl Dispatcher {
...
-            self.create_http_context(context_id, root_context_id);
-        } else if self.new_stream.get().is_some() {
-            self.create_stream_context(context_id, root_context_id);
+            self.create_root_context(context_id);
+        } else if let Some(root_context) = self.roots.borrow().get(&root_context_id) {
+            match root_context.get_type() {
+                ContextType::HttpContext => self.create_http_context(context_id, root_context_id),
+                ContextType::StreamContext => {
+                    self.create_stream_context(context_id, root_context_id)
+                }
```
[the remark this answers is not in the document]
dgn: agreed. removed it from the trait and the examples

src/traits.rs
```diff
@@ -159,6 +175,16 @@ pub trait StreamContext: Context {
     fn on_log(&mut self) {}
 }
 
+struct EmptyHttpContext;
+
+impl HttpContext for EmptyHttpContext {}
+impl Context for EmptyHttpContext {}
```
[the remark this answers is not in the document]
dgn: done

src/traits.rs
```diff
@@ -121,6 +121,22 @@ pub trait RootContext: Context {
...
+        &self,
+        _context_id: u32,
+        _root_context_id: u32,
+    ) -> Box<dyn StreamContext> {
+        Box::new(EmptyStreamContext)
+    }
+
+    fn get_type(&self) -> ContextType {
+        ContextType::RootContext
+    }
```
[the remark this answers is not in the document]
dgn: if I had to choose I'd prefer the former (ie have a single `new_context` fn that creates whatever type), but at this point I think I like the more explicit approach more- because in the end we would have to be type-casting in the SDK again to call trait-specific fn's. but I'm open to suggestions

src/dispatcher.rs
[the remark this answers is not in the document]
PiotrSikora: Sorry, I didn't notice this was called from `on_create_context()`, and I thought it was a dead code... I think it was more readable when it was split into separate functions. Could you undo this last update? Thanks and sorry for the noise.

examples/http_body.rs
```diff
@@ -18,10 +18,23 @@ use proxy_wasm::types::*;
...
+    fn get_type(&self) -> ContextType {
+        ContextType::HttpContext
+    }
+
+    fn create_http_context(&self, _context_id: u32) -> Box<dyn HttpContext> {
+        Box::new(HttpBody)
+    }
+}
+
+impl Context for HttpBodyRoot {}
```
PiotrSikora, line 37: Nit: move `impl Context` before `impl RootContext` to match order for `HttpBody`.

src/traits.rs
```diff
@@ -121,6 +121,18 @@ pub trait RootContext: Context {
...
+        Box::new(EmptyHttpContext)
+    }
+
+    fn create_stream_context(&self, _context_id: u32) -> Box<dyn StreamContext> {
+        Box::new(EmptyStreamContext)
+    }
+
+    fn get_type(&self) -> ContextType {
+        ContextType::RootContext
+    }
```
PiotrSikora, line 135: Shouldn't those functions be split into a separate trait?

examples/http_auth_random.rs
```diff
@@ -20,10 +20,23 @@ use std::time::Duration;
 #[no_mangle]
 pub fn _start() {
     proxy_wasm::set_log_level(LogLevel::Trace);
-    proxy_wasm::set_http_context(|_, _| -> Box<dyn HttpContext> { Box::new(HttpAuthRandom) });
+    proxy_wasm::set_root_context(|_| -> Box<dyn RootContext> { Box::new(HttpAuthRandomRoot) });
```
PiotrSikora, line 23: Looking at this example, I wonder if we couldn't do something type-specific here, e.g.
```
proxy_wasm::set_root_context(|_| -> Box<dyn RootContext> { Box::new(HttpHeadersRoot) });
proxy_wasm::set_http_context(|root_context, context_id| -> Box<dyn HttpContext> {
    root_context.new_http_context(context_id)
});
```
with dispatcher taking care of resolving `root_context_id` to `root_context` before calling this callback?

PiotrSikora (changes requested):

PiotrSikora (commented):
Also, could you merge/rebase on top of current master branch? Thanks!

examples/http_auth_random.rs
PiotrSikora: On the second thought, that might be useless, since we cannot expose multiple interfaces right now anyway.

src/dispatcher.rs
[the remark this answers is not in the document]
dgn: reverted

src/traits.rs
dgn: to avoid the need for a `RootContext`? I'm not sure about that. I think `RootContext` (by its name) already has a notion of controlling its child contexts, so this is what I would expect, instead of another meta-struct that creates all contexts.

Also, the main reason to implement this was so that the `on_configure` calls could be passed on to child contexts easily by the `RootContext`. if we move context creation out of the root context again, it'll again lose the ability to map `RootContext` to e.g. `HttpContext` - except if the user explicitly keeps a mapping in that struct

examples/http_auth_random.rs
```diff
@@ -20,31 +20,21 @@ use std::time::Duration;
...
 }
 
 struct HttpAuthRandom;
```
PiotrSikora, line 26: Nit: move `struct HttpAuthRandom;` after definition of `HttpAuthRandomRoot`.

examples/http_body.rs
```diff
@@ -18,10 +18,23 @@ use proxy_wasm::types::*;
 #[no_mangle]
 pub fn _start() {
     proxy_wasm::set_log_level(LogLevel::Trace);
-    proxy_wasm::set_http_context(|_, _| -> Box<dyn HttpContext> { Box::new(HttpBody) });
+    proxy_wasm::set_root_context(|_| -> Box<dyn RootContext> { Box::new(HttpBodyRoot) });
 }
 
 struct HttpBody;
```
PiotrSikora, line 24: Nit: move `struct HttpBody;` after definition of `HttpBodyRoot`.

PiotrSikora (commented):
Thanks! Could you also add an example showing developers how to use this feature? Maybe add a new header which value is passed via configuration to `examples/http_headers.rs`?

Note to self: merge this after tagging v0.1.3 (pending fix for #40).

src/types.rs
```diff
@@ -85,3 +85,9 @@ pub enum MetricType {
 }
 
 pub type Bytes = Vec<u8>;
+
+pub enum ContextType {
```
PiotrSikora, line 89: Please add:
```
#[repr(u32)]
#[derive(Debug)]
```
(it's not used in the ABI right now, but it's going to be soon, so we might as well define it properly now)

Also, could you move it below `Status`?

src/dispatcher.rs
```diff
@@ -131,26 +116,19 @@ impl Dispatcher {
...
+                ContextType::RootContext => panic!("missing ContextType on root_context"),
+            }
```
PiotrSikora, line 129: One more thing. I think it would be useful if we could retain backwards compatibility, and merge this without breaking all existing plugins, even if there aren't too many in the wild right now.

One way to do this is to fallback to the standalone constructors in case `root_context.get_type()` returns `ContextType::RootContext` or `None`, and perhaps revert changes to `examples/http_auth_random.rs` to make sure that the old code path still works.

src/traits.rs
```diff
@@ -121,6 +121,18 @@ pub trait RootContext: Context {
     fn on_queue_ready(&mut self, _queue_id: u32) {}
 
     fn on_log(&mut self) {}
+
+    fn create_http_context(&self, _context_id: u32) -> Box<dyn HttpContext> {
+        Box::new(EmptyHttpContext)
```
PiotrSikora, line 126: `Box<dyn HttpContext>` -> `Option<Box<dyn HttpContext>>` and return `None`.

src/traits.rs
```diff
@@ -121,6 +121,18 @@ pub trait RootContext: Context {
...
+    }
+
+    fn create_stream_context(&self, _context_id: u32) -> Box<dyn StreamContext> {
+        Box::new(EmptyStreamContext)
```
PiotrSikora, line 130: Ditto.

src/traits.rs
```diff
@@ -121,6 +121,18 @@ pub trait RootContext: Context {
...
+    }
+
+    fn get_type(&self) -> ContextType {
+        ContextType::RootContext
```
PiotrSikora, line 134: `Option<ContextType>` and return `None`.

src/traits.rs
```diff
@@ -159,6 +171,11 @@ pub trait StreamContext: Context {
     fn on_log(&mut self) {}
 }
 
+struct EmptyStreamContext;
+
+impl StreamContext for EmptyStreamContext {}
+impl Context for EmptyStreamContext {}
```
PiotrSikora, line 177: This won't be needed with `Option<T>`.

src/traits.rs
```diff
@@ -303,3 +320,8 @@ pub trait HttpContext: Context {
 
     fn on_log(&mut self) {}
 }
+
+struct EmptyHttpContext;
+
+impl HttpContext for EmptyHttpContext {}
+impl Context for EmptyHttpContext {}
```
PiotrSikora, line 327: This won't be needed with `Option<T>`.

PiotrSikora (changes requested):

src/dispatcher.rs
dgn: +1 to remain backwards-compatible, but I think we should then deprecate the old way to create contexts and remove it in 0.2.0

dgn: that would also allow us to remove `NoopRoot` in 0.2.0

examples/http_body.rs
dgn: done

src/types.rs
dgn: done

src/traits.rs
dgn: done

src/traits.rs
dgn: done

src/traits.rs
dgn: done

src/traits.rs
dgn: removed

examples/http_config.rs
```diff
@@ -0,0 +1,66 @@
...
+
+impl HttpContext for HttpConfigHeader {
+    fn on_http_response_headers(&mut self, _num_headers: usize) -> Action {
+        self.add_http_response_header("custom-header", self.header_content.as_str());
+
+        Action::Continue
+    }
+}
+
+struct HttpConfigHeaderRootContext {
```
PiotrSikora, line 43: Nit: name this `HttpConfigHeaderRoot` and move it before `HttpConfigHeader` to match style used in other examples.

examples/http_auth_random.rs
```diff
@@ -25,6 +25,24 @@ pub fn _start() {
...
+        }
+        trace!("Access forbidden.");
+        self.send_http_response(
+            403,
+            vec![("Powered-By", "proxy-wasm")],
+            Some(b"Access forbidden.\n"),
+        );
+    }
+}
+
```
PiotrSikora, line 45: Nit: this is unrelated and unnecessary move, please undo.

examples/http_config.rs
```diff
@@ -0,0 +1,66 @@
...
+    fn on_configure(&mut self, _plugin_configuration_size: usize) -> bool {
+        if let Some(config_bytes) = self.get_configuration() {
+            self.header_content = str::from_utf8(config_bytes.as_ref()).unwrap().to_owned()
+        }
+        true
+    }
+
+    fn create_http_context(&self, _context_id: u32) -> Option<Box<dyn HttpContext>> {
+        Some(Box::new(HttpConfigHeader {
+            header_content: self.header_content.clone(),
```
PiotrSikora, line 59: This seems expensive, since presumably each context will own a copy of the header value.

It's fine in this example, since the value is small, but in general, wouldn't it be better to store a reference to the parent context and access configuration from there?

src/dispatcher.rs
PiotrSikora: The "old way" is fine for plugins that don't have configuration (e.g. where everything is already in the code), and the "new way" can double the code size with unnecessary boilerplate.

But it's something to consider for sure.

src/types.rs
```diff
@@ -47,6 +47,13 @@ pub enum Status {
     InternalFailure = 10,
 }
 
+#[repr(u32)]
+#[derive(Debug)]
+pub enum ContextType {
+    HttpContext = 0,
```
PiotrSikora, line 53: Could you re-add `RootContext` as `0` here? Thanks.

dgn: Why? How would that be different from returning `None`?

PiotrSikora (approved):

PR #42, opened 2020-10-26
Allow building for wasm32-wasi target.

PiotrSikora:

### DIFF
--- a/src/allocator.rs
+++ b/src/allocator.rs
@@ -15,8 +15,12 @@
 #[global_allocator]
 static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
 
+#[cfg_attr(
+    all(target_arch = "wasm32", target_os = "unknown"),
+    export_name = "malloc"
+)]
 #[no_mangle]
-pub extern "C" fn malloc(size: usize) -> *mut u8 {
+pub extern "C" fn proxy_on_memory_allocate(size: usize) -> *mut u8 {
     let mut vec: Vec<u8> = Vec::with_capacity(size);
     unsafe {
         vec.set_len(size);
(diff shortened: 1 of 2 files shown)


src/allocator.rs
```diff
@@ -15,8 +15,12 @@
 #[global_allocator]
 static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
 
+#[cfg_attr(
+    all(target_arch = "wasm32", target_os = "unknown"),
+    export_name = "malloc"
+)]
 #[no_mangle]
-pub extern "C" fn malloc(size: usize) -> *mut u8 {
+pub extern "C" fn proxy_on_memory_allocate(size: usize) -> *mut u8 {
```
gbrail, line 23: I don't recall which version of the ABI this is in -- if we change this now will it affect compatibility? Or do we need to bump the ABI version?

gbrail (commented):
This makes sense but I'm just not sure whether we need to do anything about ABI compatibility if we want to use this function instead of malloc.

src/allocator.rs
PiotrSikora: It's part of the updated-not-yet-finalized ABI (https://github.com/proxy-wasm/spec/pull/1), and it's already enabled as a fallback in Envoy (https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/80).

Note that this is still exported as `malloc` when using `wasm32-unknown-unknown` target, and it's only exported as `proxy_on_memory_allocate` when using `wasm32-wasi` and other non-WebAssembly targets, which previously were impossible to use at all, so we're not breaking any existing compatibility.

It's not perfect, but it's the best we can do without making huge changes (and I'd rather avoid those now, since I want to finalize the ABI over the next week or two, and them upgrade the SDK straight to that version).

gbrail (approved):
I get it -- there is a usage of "cfg_attr" that I didn't know about. Makes sense!

PR #49, opened 2020-11-05
Add an example to add custom metrics on proxy wasm

alandiegosantos:


examples/http_metrics.rs
```diff
@@ -0,0 +1,99 @@
...
+pub trait CounterTrait {
+    /// Increments counter by `1`.
+    fn inc(&self) -> Result<(), Status> {
+        self.add(1)
+    }
+    /// Increments counter by a given offset.
+    fn add(&self, offset: u64) -> Result<(), Status>;
+    /// Returns current value of the counter.
+    fn value(&self) -> Result<u64, Status>;
+}
```
PiotrSikora, line 94: Should this trait (with default implementation from above) be added to `src/traits.rs` instead?

PiotrSikora (commented):
You need to add those examples to `Cargo.toml` to pass the CI.

examples/http_metrics.rs
```diff
@@ -0,0 +1,99 @@
...
+
+fn counter(name: &str) -> Result<Box<dyn CounterTrait>, Status> {
+    hostcalls::define_metric(MetricType::Counter, name)
+        .map(|metric_id| Box::new(Counter(metric_id)) as Box<dyn CounterTrait>)
+}
```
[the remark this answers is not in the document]
alandiegosantos: Envoy is smart in this topic and increment the right metric. I agree with moving this to the startup. I will take a look at it later.

examples/http_metrics.rs
alandiegosantos: `http_metrics.rs `was supposed to be just an example of how to use metrics API. 
CounterTrait can be moved to `src/traits.rs` along with Gauge and Histogram implementations, adding some abstraction to the code.
However, should this lib be as close as possible from proxy-wasm spec?

PR #50, opened 2020-11-06
Allow building for wasm32-wasi target using Bazel.

PiotrSikora:
While there, update Bazel to 3.7.0.


.bazelrc
```diff
@@ -1,2 +1,2 @@
-# Force Bazel to use --target=wasm32-unknown-unknown.
-build --platforms=@io_bazel_rules_rust//rust/platform:wasm
+# Force Bazel to use --target=wasm32-wasi
+build --platforms=@io_bazel_rules_rust//rust/platform:wasi
```
gbrail, line 2: Are we ready to change this by default? I've only ever used Rust and Wasm with the "wasm32-unknown-unknown" target and I'm not sure how well WASI works.

gbrail (approved):
This is fine but are we sure we're ready to change the default target in .bazelrc?

.bazelrc
PiotrSikora: Proxy-Wasm C++ SDK (Emscripten) already emits WASI calls, and we have the basic ones implemented in Envoy already (e.g. `wasi_args_get`, `wasi_environ_get`, `wasi_fd_write`, `wasi_proc_exit`, and more recently [`wasi_clock_get_time`](https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/65) and [`wasi_random_get`](https://github.com/proxy-wasm/proxy-wasm-cpp-host/pull/68)).

We're not losing any functionality by switching to it, and only gaining access to features exposed by WASI-capable crates (see: https://github.com/proxy-wasm/proxy-wasm-rust-sdk/pull/51).

PiotrSikora: Does the change to default make sense for you?

gbrail: That makes sense to me -- when I built locally I also got an import of some super long mangled name, but perhaps thats' changed. As long as WASI builds work with a recent Envoy then I think this change is fine.

PR #53, opened 2020-11-10
Use cargo-raze's gen_buildrs for trusted crates.

PiotrSikora:


gbrail (approved):
That makes Bazel and Rust make a lot more sense -- LGTM!

PR #56, opened 2020-11-24
Add support for nested HTTP callouts.

### LINKED ISSUE #40: It's not possible to issue a second http call
Hi,

Currently it's not possible to issue another http call from `Context::on_http_call_response`. Is this intentional ? If it is how is one supposed to make several http calls, for example fetch an oauth token, call another service with it, then resume the request ?

The error is:

```
[2020-10-25 11:04:04.792][12][critical][wasm] [external/envoy/source/extensions/common/wasm/context.cc:1018] wasm log: panicked at 'already borrowed: BorrowMutError', /Users/svetlin/.cargo/registry/src/github.com-1ecc6299db9ec823/proxy-wasm-0.1.2/src/dispatcher.rs:137:14
```

I believe the error is caused because the dispatcher [tries to borrow the callouts for a second time ](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/e7a6ad0f123965b5c05fa85606da9da00cac6ab6/src/dispatcher.rs#L137) when registering the second http call. 

The [first borrow occurs when calling the on_http_call_response callback](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/e7a6ad0f123965b5c05fa85606da9da00cac6ab6/src/dispatcher.rs#L348). The duration of the borrow lasts for the whole `if let Some()` body.

Given that the context_id is `Copy`, this method can be modified like that:

```
{
        let context_id = match self.callouts.borrow_mut().remove(&token_id) {
            Some(id) => id,
            None => panic!("invalid token_id")
        };

        if let Some(http_stream) = self.http_streams.borrow_mut().get_mut(&context_id) {
            self.active_id.set(context_id);
            hostcalls::set_effective_context(context_id).unwrap();
            http_stream.on_http_call_response(token_id, num_headers, body_size, num_trailers)
        } else if let Some(stream) = self.streams.borrow_mut().get_mut(&context_id) {
            self.active_id.set(context_id);
            hostcalls::set_effective_context(context_id).unwrap();
            stream.on_http_call_response(token_id, num_headers, body_size, num_trailers)
        } else if let Some(root) = self.roots.borrow_mut().get_mut(&context_id) {
            self.active_id.set(context_id);
            hostcalls::set_effective_context(context_id).unwrap();
            root.on_http_call_response(token_id, num_headers, body_size, num_trailers)
        }
    }
```

Thus eliminating the borrow on the `callouts` member, which resolves the issue.

What do you think ?

SvetlinZarev-SAP:
Fixes #40.

### DIFF
--- a/src/dispatcher.rs
+++ b/src/dispatcher.rs
@@ -345,22 +345,24 @@ impl Dispatcher {
         body_size: usize,
         num_trailers: usize,
     ) {
-        if let Some(context_id) = self.callouts.borrow_mut().remove(&token_id) {
-            if let Some(http_stream) = self.http_streams.borrow_mut().get_mut(&context_id) {
-                self.active_id.set(context_id);
-                hostcalls::set_effective_context(context_id).unwrap();
-                http_stream.on_http_call_response(token_id, num_headers, body_size, num_trailers)
-            } else if let Some(stream) = self.streams.borrow_mut().get_mut(&context_id) {
-                self.active_id.set(context_id);
-                hostcalls::set_effective_context(context_id).unwrap();
-                stream.on_http_call_response(token_id, num_headers, body_size, num_trailers)
-            } else if let Some(root) = self.roots.borrow_mut().get_mut(&context_id) {
-                self.active_id.set(context_id);
-                hostcalls::set_effective_context(context_id).unwrap();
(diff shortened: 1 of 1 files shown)


PiotrSikora (approved):

PiotrSikora:
Thanks!

PR #67, opened 2020-12-23
Introduce on_create_child_context hook for creating non root contexts

unleashed:
This is based on previous work from @yskopets for creating child contexts via a hook in the root context (#6).

The patches first introduce the mechanism, then change the examples to make use of it, and finally remove the other currently existing methods in a **breaking** change to both simplify usage by having a single way to do this and clean up some of the code which should no longer be necessary. Compile-tested all examples after every patch.

Feel free to use or cherry pick partially or as needed.

### DIFF
--- a/examples/http_auth_random.rs
+++ b/examples/http_auth_random.rs
@@ -20,7 +20,17 @@ use std::time::Duration;
 #[no_mangle]
 pub fn _start() {
     proxy_wasm::set_log_level(LogLevel::Trace);
-    proxy_wasm::set_http_context(|_, _| -> Box<dyn HttpContext> { Box::new(HttpAuthRandom) });
+    proxy_wasm::set_root_context(|_| -> Box<dyn RootContext> { Box::new(HttpAuthRandomRoot) });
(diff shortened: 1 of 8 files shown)


unleashed:
@googlebot I signed it!

PiotrSikora:
Hi @unleashed! Sorry, I've missed the email notifications for your PRs.

Is this only a "style rewrite" of what was already implemented in #34, or am I missing something? In any case, while I agree that this is a nicer interface, the next version of the ABI is most likely going to remove generic context type, and plugins will be able to support multiple extensions points (e.g. both HTTP and TCP), so we're going to need typed constructors, which this PR removes.

unleashed:
Yes, this is a rewrite of #34 with a better user development experience and slightly nicer internal structure.

So if we'll need typed constructors anyway it's not worth it to merge this, but I'd like to take a look and see how we could improve on the current interface while allowing the incoming changes. Do you have a link to these changes and/or to a place where the ABI evolution is being discussed?

PR #68, opened 2020-12-23
logger: don't panic in the panic handler

unleashed:
This avoids a potential double-panic in the panic handler if the logging status code is not recognised.

### DIFF
--- a/src/hostcalls.rs
+++ b/src/hostcalls.rs
@@ -21,12 +21,20 @@ extern "C" {
     fn proxy_log(level: LogLevel, message_data: *const u8, message_size: usize) -> Status;
 }
 
+// Useful for the panic hook to avoid a double-panic.
+#[inline]
+pub(crate) fn log_nopanic(level: LogLevel, message: &str) -> Result<Status, Status> {
+    let status = unsafe { proxy_log(level, message.as_ptr(), message.len()) };
+    match status {
+        ok @ Status::Ok => Ok(ok),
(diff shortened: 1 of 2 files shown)


src/hostcalls.rs
```diff
@@ -21,12 +21,20 @@ extern "C" {
     fn proxy_log(level: LogLevel, message_data: *const u8, message_size: usize) -> Status;
 }
 
+// Useful for the panic hook to avoid a double-panic.
+#[inline]
+pub(crate) fn log_nopanic(level: LogLevel, message: &str) -> Result<Status, Status> {
```
PiotrSikora, line 26: Is there any reason for using `Result<Status, Status>` instead of `Result<(), Status>` like rest of the SDK?

PiotrSikora (commented):
This is technically correct (good catch!), but looking at the code, I cannot see how `proxy_log` could panic here. Did you run into this double-panic?

src/hostcalls.rs
unleashed: Just leaving the decision about what to do with the `Status` to the callers out of habit. Since this fn is private to this crate it won't show as part of the SDK to users, and this allows for callers to have more flexibility to make that decision or be extended based on this value without needing to modify this fn.

unleashed:
No, I didn't run into it. I just wanted to protect against the possibility that changes happen elsewhere and a new code path opens inadvertently that allows this to happen.

PR #73, opened 2021-02-03
Check only for outdated root dependencies in Cargo checks.

nullpo-head:
This PR makes `cargo outdated` check only direct dependencies only in CI.
Currently, CI also checks the versions of all dependencies in the dependency chain of the crates that `proxy-wasm-rust-sdk` depends on, and fails if it finds an outdated one. However, they are not direct dependencies of `proxy-wasm-rust-sdk` and cannot be updated by its Cargo.toml.

The current result of `cargo outdated`. 
```
Name             Project                        Compat  Latest                         Kind    Platform
----             -------                        ------  ------                         ----    --------
chrono->time     0.1.43                         ---     0.1.44                         Normal  ---
getrandom->wasi  0.10.2+wasi-snapshot-preview1  ---     0.10.0+wasi-snapshot-preview1  Normal  cfg(target_os = "wasi")
```

That of `cargo outdated --root-deps-only`
```
All dependencies are up to date, yay!
```

This PR also includes dependency updates.


PiotrSikora:
Thanks! I have similar change in my local tree, so this looks good, but you need to sign the CLA before I can approve this.

Also, could you please split this into 2 separate PRs (one with `--root-deps-only`, and another with updates to dependencies)? Those are 2 different changes in my mind.

PiotrSikora:
Thanks for signing the CLA! Could you open the other PR as well? Thanks!

nullpo-head:
@googlebot I signed it!

nullpo-head:
Thanks so much for your review, Piotr! I've split this PR and opened another one. Although `outdated` CI is failing for this PR, the other PR to update dependencies will fix it.

PiotrSikora (approved):
Thanks!

PR #77, opened 2021-02-07
[proposal] Extend the run to bazel 3.7.*

Gsantomaggio:
Bazel 3.7.2 is out.
This PR extends the build to Bazel 3.7.*.
Thank you


Gsantomaggio:
never mind, it seems that bazelisk doesn't build with `3.7.*`.
That's strange since:
```
bazel --bazelrc=/dev/null build --platforms=@rules_rust//rust/platform:wasm //...
```
works

Issue #82, opened 2021-02-25
Error in hostcalls functions causes filter to panic


alandiegosantos:
## Summary
The hostcalls panic if the status is not expected. It causes filter to stop working if there is any non-expected status response in any ABI call.
```
[2021-02-24 23:03:59.403][25254][critical][wasm] [external/envoy/source/extensions/common/wasm/context.cc:1157] wasm log: panicked at 'unexpected status: 2', external/raze__proxy_wasm__0_1_3/src/hostcalls.rs:233:23
```

## Expectation
The Status error is propagated through the context calls, so each dev can deal with the error.

alandiegosantos:
#83

PiotrSikora:
How are you getting this panic and what proxy (and version) are you using? Do you have a simple reproducer?

If both proxy and SDK implement the same ABI correctly, then there are no unexpected status codes, and you should never see that panic, but clearly there is something off here.

Note that _expected_ errors are propagated through context call.

alandiegosantos:
The panic was observed on Envoy 1.17 on OSX when the client disconnects while waiting for the response.
```
$ envoy --version
envoy  version: 5c801b25cae04f06bf48248c90e87d623d7a6283/1.17.0/Modified/DEBUG/BoringSSL
```
And, this is the *on_log* fn:
```
impl HttpContext for PanicFilter {
    fn on_log(&mut self) {
        if self.get_http_response_header(":status").is_none() {
            error!("Client disconnected while waiting for the response");
        }
    }
}
```
And these are the generated logs.
```
[2021-02-27 00:08:58.637][62818][critical][wasm] [source/extensions/common/wasm/context.cc:1180] wasm log: panicked at 'unexpected status: 2', external/raze__proxy_wasm__0_1_3/src/hostcalls.rs:233:23
[2021-02-27 00:08:58.640][62818][error][wasm] [source/extensions/common/wasm/wasm_vm.cc:39] Function: proxy_on_log failed: Uncaught RuntimeError: unreachable
Proxy-Wasm plugin in-VM backtrace:
  0:  0x11446 - rust_panic
  1:  0x111e8 - _ZN3std9panicking20rust_panic_with_hook17h4f753dc70b771d8eE
  2:  0x11680 - _ZN3std9panicking19begin_panic_handler28_$u7b$$u7b$closure$u7d$$u7d$17h5c9fc29957309ae8E
[...]

PR #83, opened 2021-02-25
Propagate the non-OK status instead of panic on hostcalls

alandiegosantos:
Replaced panic by Err in Result.

### DIFF
--- a/src/hostcalls.rs
+++ b/src/hostcalls.rs
@@ -25,7 +25,7 @@ pub fn log(level: LogLevel, message: &str) -> Result<(), Status> {
     unsafe {
         match proxy_log(level, message.as_ptr(), message.len()) {
             Status::Ok => Ok(()),
-            status => panic!("unexpected status: {}", status as u32),
+            status => Err(status),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


alandiegosantos:
IMO, we can remove the _unwrap_ from the functions in the _trait.rs_, but this would cause a breaking change in the API. 

It is also possible to add another set of functions that returns a Result, but it will duplicate the code.

@PiotrSikora Do you have an opinion about that?

PiotrSikora:
I agree that libraries shouldn't exit or panic, and instead they should propagate the error to the caller. As mentioned in #82, valid errors are propagated from the hostcalls where they are expected.

However, panics that you're replacing in this PR can only happen if host environment and/or SDK don't implement the ABI correctly, i.e. only after receiving an unexpected error that cannot happen per specification, which is closer to an undefined behavior than a runtime error. This has been invalueable in finding errors across implementations, so I'd prefer to leave the code as-is... at least for the time being.

alandiegosantos:
I agree with you. 
I was not aware that #82 was caused by a problem with ABI on Envoy.
Closing this PR.

PR #92, opened 2021-03-15
Update rules_rust to latest.

PiotrSikora:

### DIFF
--- a/bazel/repositories.bzl
+++ b/bazel/repositories.bzl
@@ -19,7 +19,7 @@ def proxy_wasm_rust_sdk_repositories():
     maybe(
         http_archive,
         name = "rules_rust",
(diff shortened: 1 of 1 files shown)


mathetake:
I see that now `rust.bzl` is deprecated and we are recommended to use `def.bzl`.

PiotrSikora:
> I see that now `rust.bzl` is deprecated and we are recommended to use `defs.bzl`.
> 
> https://github.com/bazelbuild/rules_rust/blob/main/rust/rust.bzl#L15

Thanks for noticing this! Unfortunately, `cargo-raze` doesn't support it yet (see: https://github.com/google/cargo-raze/issues/398), so I'd prefer to wait for it, instead of having mixed use of `rust.bzl` and `defs.bzl` in the same repository.

mathetake (approved):
> so I'd prefer to wait for it, instead of having mixed use of rust.bzl and defs.bzl in the same repository.

OK, that makes sense. LGTM!

PR #95, opened 2021-03-19
Update libc (transitive dependency) to v0.2.90.

PiotrSikora:

### DIFF
--- a/bazel/cargo/Cargo.raze.lock
+++ b/bazel/cargo/Cargo.raze.lock
@@ -64,9 +64,9 @@ dependencies = [
 
 [[package]]
 name = "libc"
-version = "0.2.89"
+version = "0.2.90"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "538c092e5586f4cdd7dd8078c4a79220e3e168880218124dcbce860f0ea938c6"
+checksum = "ba4aede83fc3617411dc6993bc8c70919750c1c257c6ca6a502aed6e0e2394ae"
 
 [[package]]
 name = "log"
--- a/bazel/cargo/crates.bzl
+++ b/bazel/cargo/crates.bzl
@@ -73,12 +73,12 @@ def raze_fetch_remote_crates():
 
     maybe(
         http_archive,
-        name = "raze__libc__0_2_89",
-        url = "https://crates.io/api/v1/crates/libc/0.2.89/download",
+        name = "raze__libc__0_2_90",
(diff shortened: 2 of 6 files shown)


mathetake (approved):
LGTM (I'm wondering if we could automate the transitive dependency update? e.g. adding scheduled github action to create PRs like this)

PiotrSikora:
> LGTM (I'm wondering if we could automate the transitive dependency update? e.g. adding scheduled github action to create PRs like this)

We could... Alternatively, we could ignore transitive dependencies, since we need them only for Bazel builds (we don't ship `Cargo.lock`, so Cargo builds are always using the latest versions), and even then, most of them are used only in examples:
```
$ cargo tree --target=wasm32-wasi --edges normal
proxy-wasm v0.1.3 (/path/to/proxy-wasm-rust-sdk)
├── hashbrown v0.11.0
│   └── ahash v0.7.2
│       ├── getrandom v0.2.2
│       │   ├── cfg-if v1.0.0
│       │   └── wasi v0.10.2+wasi-snapshot-preview1
│       └── once_cell v1.7.2
└── log v0.4.14
    └── cfg-if v1.0.0
```

PiotrSikora:
> > LGTM (I'm wondering if we could automate the transitive dependency update? e.g. adding scheduled github action to create PRs like this)
> 
> We could... Alternatively, we could ignore transitive dependencies, since we need them only for Bazel builds (we don't ship `Cargo.lock`, so Cargo builds are always using the latest versions), and even then, most of them are used only in examples:
> 
> ```
> $ cargo tree --target=wasm32-wasi --edges normal
> proxy-wasm v0.1.3 (/path/to/proxy-wasm-rust-sdk)
> ├── hashbrown v0.11.0
> │   └── ahash v0.7.2
> │       ├── getrandom v0.2.2
> │       │   ├── cfg-if v1.0.0
> │       │   └── wasi v0.10.2+wasi-snapshot-preview1
> │       └── once_cell v1.7.2
> └── log v0.4.14
>     └── cfg-if v1.0.0
> ```

@mathetake any thoughts on this?

mathetake:
Sorry forgot to reply. If we could assume that majority of users would use cargo, not Basel, then let’s ignore the transitive dependencies upgrades.

codefromthecrypt:
I agree. the bazel part is distracting and personally not helpful even if some like it.

PiotrSikora:
@codefromthecrypt yup, the freshness check was removed in #99, and I'm going to update Bazel deps only before creating a new release.

PR #99, opened 2021-03-29
Remove freshness check for Cargo.raze.lock.

PiotrSikora:


mathetake (approved):
Thanks!

PR #124, opened 2021-09-16
Allow arbitrary bytes as upstream name of grpc call

krdln:
Envoy allows the upstream to be serialized GrpcService message, but the proxy-wasm API allows only `&str`. (why use GrpcService message – to be able to connect to arbitrary upstream not predefined in envoy's clusters). This functionality seems to be supported by envoy so it makes sense to expose it in SDK.

This is obviously a breaking change, not sure how to make it non-breaking though (I thought about `imp AsRef<u8>`, but that would break object safety, I guess. Perhaps a separate trait method would a better solution?)

I've also looked a bit in the [spec](https://github.com/proxy-wasm/spec/tree/master/abi-versions/vNEXT#proxy_dispatch_grpc_call), in which the `upstream` argument is called `grpc_service` (hinting the use of formentioned GrpcService message), but in [your PR](https://github.com/proxy-wasm/spec/pull/1#discussion_r473200070), this argument is called just `upstream_name`. So thus I'm confused on the state of that feature, can you shed some light there?


PiotrSikora:
> I've also looked a bit in the [spec](https://github.com/proxy-wasm/spec/tree/master/abi-versions/vNEXT#proxy_dispatch_grpc_call), in which the `upstream` argument is called `grpc_service` (hinting the use of formentioned GrpcService message), but in [your PR](https://github.com/proxy-wasm/spec/pull/1#discussion_r473200070), this argument is called just `upstream_name`. So thus I'm confused on the state of that feature, can you shed some light there?

IMHO, the ability to declare upstreams in Envoy's implementation was a mistake and should be removed (but that's a bit tricky, since Istio Telemetry plugins are using it), since it breaks sandboxing in a pretty fundamental way.

Is there any specific reason why you want/need this feature?

PR #125, opened 2021-09-16
Make map serialization/deserialization utils public 

krdln:
(see commit messages)


PiotrSikora:
Hi @krdln! Could you check the CLA? It looks that the email is on file, but your GitHub username is not.

PR #130, opened 2022-01-26
Fix lint warning on memory allocation

Swaagie:
Fix clippy lint warning to safely initialize potentially unallocated memory with MaybeUninit


src/allocator.rs
```diff
@@ -16,13 +16,15 @@
 #[global_allocator]
 static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
 
+use std::mem::MaybeUninit;
```
PiotrSikora, line 19: Style: `use` before global variables.

Swaagie: 👍 oversight on my end

PiotrSikora (approved):
Thanks!

PR #129, opened 2022-01-26
Update dependencies and resolve workflow failures

Swaagie:
Thought this might be a meaningful first contribution, updated dependencies and opted out of using the `chrono` crate for the `hello_world` example. I have no experience with bazel/raze before (up until now) so if I'm making crazy changes please let me know. 

- Updated `cargo-raze` to 0.14.1 (and adjusted build configs)
- Updated to use rust edition 2021 as `cargo metadata` depended on latest without having to use nightly.
- Removed `chrono` crate in favor of `time` crate to avoid having to wait [for a resolution for the sec vun](https://github.com/chronotope/chrono/pull/578#issuecomment-974493484)
- Updated `rust_rules` to [b3ddf6f096887b757ab1a661662a95d6b2699fa7](https://github.com/bazelbuild/rules_rust/commit/b3ddf6f096887b757ab1a661662a95d6b2699fa7)
- Fixed clippy suggested safe read of potentially unallocated memory with `MaybeUninit`
- Regenerated lockfiles

Happy to change and update the PR where needed.

PR #131, opened 2022-01-26
Use time crate over chrono for hello_world example

Swaagie:
Sidestep reported [security vulnerability for which there is no clear resolution](https://github.com/chronotope/chrono/pull/578#issuecomment-974493484). 

Extracted from: #129


PiotrSikora:
Could you split `hashbrown` update from the `chrono`/`time` swap? Thanks!

PR #132, opened 2022-01-26
Update cargo-raze to v0.14.1.

Swaagie:
Extracted from: #129

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,6 @@
 
 When updating dependencies, you need to regenerate Bazel `BUILD` files to match updated `Cargo.toml`:
 ```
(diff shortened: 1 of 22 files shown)


PiotrSikora:
You need to regenerate artifacts.

Swaagie:
Done, was hoping to avoid rebase, considering https://github.com/proxy-wasm/proxy-wasm-rust-sdk/pull/131 also has regenerated.

PiotrSikora:
> Done, was hoping to avoid rebase, considering #131 also has regenerated.

IMHO, this is better, otherwise #131 would mix updates with `chrono`/`time` swap, which would make it less readable.

PR #133, opened 2022-01-26
Update rules_rust to latest (with Rust v1.58.1).

Swaagie:
This will likely require the lockfiles to be regenerated to update `edition` versions, but to prevent conflicts it's better to do this after https://github.com/proxy-wasm/proxy-wasm-rust-sdk/pull/131 has been merged.

Extracted from: #129

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -3,7 +3,7 @@ load("@rules_rust//rust:defs.bzl", "rust_library")
 rust_library(
     name = "proxy_wasm",
     srcs = glob(["src/*.rs"]),
-    edition = "2021",
+    edition = "2018",
     visibility = ["//visibility:public"],
     deps = [
         "//bazel/cargo:hashbrown",
--- a/BUILD
+++ b/BUILD
@@ -1,9 +1,9 @@
-load("@rules_rust//rust:rust.bzl", "rust_library")
+load("@rules_rust//rust:defs.bzl", "rust_library")
 
 rust_library(
     name = "proxy_wasm",
     srcs = glob(["src/*.rs"]),
-    edition = "2018",
+    edition = "2021",
     visibility = ["//visibility:public"],
     deps = [
         "//bazel/cargo:hashbrown",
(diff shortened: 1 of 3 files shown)


BUILD
```diff
@@ -1,9 +1,9 @@
-load("@rules_rust//rust:rust.bzl", "rust_library")
+load("@rules_rust//rust:defs.bzl", "rust_library")
 
 rust_library(
     name = "proxy_wasm",
     srcs = glob(["src/*.rs"]),
-    edition = "2018",
+    edition = "2021",
```
PiotrSikora, line 6: Is there any particular reason why we want/need to update `edition` as part of this update?

Swaagie: Kept running into a `cargo features` error which required latest, that said considering I did all the work in one big chunk last time it was unclear whether that error orginated from cargo-raze or the rust_rules it was working with or a combination of the two. 

```
Caused by:
    0: Failed to fetch Metadata with `/home/runner/.cache/bazel/_bazel_runner/30c969fa2169879457b0b953e2c8838b/execroot/proxy_wasm_rust_sdk/bazel-out/k8-fastbuild/bin/external/cargo_raze/tools/raze.runfiles/proxy_wasm_rust_sdk/external/rust_linux_x86_64/bin/cargo` from `/home/runner/work/proxy-wasm-rust-sdk/proxy-wasm-rust-sdk`
    1: `cargo metadata` exited with an error: error: failed to parse manifest at `/home/runner/work/proxy-wasm-rust-sdk/proxy-wasm-rust-sdk/Cargo.toml`
       
       Caused by:
         feature `edition2021` is required
       
         this Cargo does not support nightly features, but if you
         switch to nightly channel you can add
         `cargo-features = ["edition2021"]` to enable this feature
```

Swaagie: GHA run https://github.com/Swaagie/proxy-wasm-rust-sdk/runs/4949540327?check_suite_focus=true

PiotrSikora: I cannot repliacte this issue locally, and I didn't run into it in other repos where I updated to this exact version of `rules_rust` (although, I'm not using `cargo raze` via Bazel there).

Could we drop `edition` and see what happens after merge with `master`?

PiotrSikora: Looks like it works fine without the `edition` bump.

PR #135, opened 2022-01-31
Update hashbrown to v0.12.

Swaagie:
- Update to hashbrown@0.12
- Regenerated lockfile

Extracted from: #129


bazel/cargo/remote/BUILD.hashbrown-0.12.0.bazel
```diff
@@ -43,7 +45,7 @@ rust_library(
     ],
     crate_root = "src/lib.rs",
     data = [],
-    edition = "2018",
+    edition = "2021",
```
PiotrSikora, line 48: This breaks CI.

FWIW, I find it easier to regenerate cargo raze artifacts than try to resolve merge/rebase failures.

Swaagie: I see, this likely led me to updating to 2021

The 2021 is brought in by a lockfile regenerated in this case, do we still want to go forward with updating to 2021 across the board?

Just experienced the rebase/merge woes with accidentally closing https://github.com/proxy-wasm/proxy-wasm-rust-sdk/pull/136 :)

PiotrSikora: Ah, it looks that `hashbrown` updated the edition: https://github.com/rust-lang/hashbrown/commit/6cbdd5a0646c108205d41e492f3106901237a035 (for no apparent reason?).

Let's see if it works now that we have Rust 1.58.1.

PiotrSikora: Yeah, it works fine.

PiotrSikora: I don't think we need the update to 2021, but it's now kind of enforced by the use of `hashbrown`. Hm.

Swaagie: Hmm yeah, considering https://blog.rust-lang.org/inside-rust/2021/03/04/planning-rust-2021.html highlights it mostly about setting up Rust for a consistent release cadence there is likely not much harm from a technical perspective, unsure how many are extending proxy-wasm and would require to follow suit. 

Alternative would be having `outdated` ignore hashbrown. Don't know if that is possible though

Swaagie: `cargo outdated -i hashbrown` is an alternative option to delay updating to 2021 to a later point in time.

PiotrSikora: It's unfortunate, but let's merge it. It looks that there is a decent number of commits since the previous version.

PR #136, opened 2022-01-31
Use time in stead of chrono in hello world example

Swaagie:
Rust/audit ✅ by sidestepping reported [security vulnerability for which there is no clear resolution](https://github.com/chronotope/chrono/pull/578#issuecomment-974493484). 
Regenerated lockfiles

Extracted from: #129


Cargo.toml
```diff
@@ -18,7 +18,7 @@ wee_alloc = { version = "0.4", optional = true }
 
 [dev-dependencies]
 cfg-if = "1.0"
-chrono = "0.4"
+time = "0.3.7"
```
PiotrSikora, line 21: Nit: There is no reason to pin to a specific version of this crate, use `0.3` instead.

examples/hello_world.rs
```diff
@@ -42,11 +42,11 @@ impl RootContext for HelloWorld {
     fn on_tick(&mut self) {
         cfg_if! {
             if #[cfg(all(target_arch = "wasm32", target_os = "unknown"))] {
-                let now: DateTime<Utc> = self.get_current_time().into();
+                let now: OffsetDateTime = OffsetDateTime::now_utc();
```
PiotrSikora, line 45: This isn't equivalent replacement, and most likely doesn't work at all, since there is no `std::time::SystemTime` for the `wasm32-unknown-unknown` platform.

I think this should be `OffsetDateTime::from_unix_timestamp_nanos(self.get_current_time());`

Swaagie: Good catch, hadn't consider that. Updated to use `get_current_time`. Verified local with with `wasm32-unknown-unknown` build target.

```
example_envoy_1 | [2022-01-31 13:35:41.634][1][debug][upstream] [source/common/upstream/strict_dns_cluster.cc:178] DNS refresh rate reset for echo_service, refresh rate 5000 ms
example_envoy_1 | [2022-01-31 13:35:45.770][19][info][wasm] [source/extensions/common/wasm/context.cc:1172] wasm log rbi_filter_id rbi_injection_vm_id: It's 2022-01-31 13:35:45.770877 +00:00:00, there is no lucky number.
```

Swaagie:
Is this good to merge?

jcrugzz (approved):
Would be cool to get this merged just to clear up the red X on CI. LGTM

PiotrSikora (changes requested):
IMHO, this change is wrong. The 2 CVEs affecting `chrono` are:

- [RUSTSEC-2020-0071](https://rustsec.org/advisories/RUSTSEC-2020-0071) - which can be easily fixed by removing `oldtime` feature from `chrono` dependency.

- [RUSTSEC-2020-0159](https://rustsec.org/advisories/RUSTSEC-2020-0159) - which doesn't have a clear fix, but it also doesn't affect this project. The CVE is about the unsafe use of `localtime_r` syscall, which isn't exposed in `wasm32-*` targets, nor used with `DateTime<Utc>`. Futhermore, the `time` crate didn't solve the underlying issue, but they've hidden calls to `localtime_r` behind a feature flag, so I don't think it's much better, security-wise. Lastly, I believe that `chrono` is (or was?) more popular in the Rust ecosystem, so I'd prefer to leave it as a more representative example. I agree that we should fix the `cargo-audit` warning, but that's going to be cleaned up with #140... in the meantime, we can exclude `RUSTSEC-2020-0159` using `--ignore` flag.

Swaagie:
Closing in favor of https://github.com/proxy-wasm/proxy-wasm-rust-sdk/pull/144

PR #144, opened 2022-04-06
Address reported RUSTSEC vulnerabilities.

### LINKED ISSUE #142: Address RUSTSEC-2020-0071 and RUSTSEC-2020-0159
From https://github.com/proxy-wasm/proxy-wasm-rust-sdk/pull/136#pullrequestreview-931084035 (for tracking purposes, in case it won't be fixed by that PR):

> IMHO, this change is wrong. The 2 CVEs affecting `chrono` are:
> 
> * [RUSTSEC-2020-0071](https://rustsec.org/advisories/RUSTSEC-2020-0071) - which can be easily fixed by removing `oldtime` feature from `chrono` dependency.
> * [RUSTSEC-2020-0159](https://rustsec.org/advisories/RUSTSEC-2020-0159) - which doesn't have a clear fix, but it also doesn't affect this project. The CVE is about the unsafe use of `localtime_r` syscall, which isn't exposed in `wasm32-*` targets, nor used with `DateTime<Utc>`. Futhermore, the `time` crate didn't solve the underlying issue, but they've hidden calls to `localtime_r` behind a feature flag, so I don't think it's much better, security-wise. Lastly, I believe that `chrono` is (or was?) more popular in the Rust ecosystem, so I'd prefer to leave it as a more representative example. I agree that we should fix the `cargo-audit` warning, but that's going to be cleaned up with #140... in the meantime, we can exclude `RUSTSEC-2020-0159` using `--ignore` flag.

cc @Swaagie

Swaagie:
- Ignore RUSTSEC-2020-0159 for now
- Use limited set of features from Chrono to address RUSTSEC-2020-0071  
- Regenerate lockfile

fixes #142


PiotrSikora:
Could we avoid the noise of regenerating the lockfile? You could manually remove `time` from there and re-run `cargo-raze` without any arguments.

PiotrSikora:
Alternatively, we could pull this after #145 to avoid the unrelated updates in this PR.

Swaagie:
Updated to removal of time only.

PiotrSikora (approved):
Thanks!

PR #146, opened 2022-04-07
Move examples into separate folders.

### LINKED ISSUE #35: WASM module built through Proxy-wasm RUST SDK can't be loaded due to missing imports: due to a missing import: __wbindgen_externref_xform__.__wbindgen_externref_table_grow
Reproduce steps:
1. Git clone the following WASM/RUST sample project:
 https://github.com/otomato-gh/proxy-wasm-rust

2. In the Cargo.toml file, add a 'reqwest' depedency
[dependencies]
reqwest = { version = "0.10" }

3. In src/lib.rs file, add:
use reqwest::Client;

and in the on_http_request_headers(...) method body:
  let _client = Client::new();

4. Build WASM using: 
cargo build --target=wasm32-unknown-unknown --release

5. Start Istio-proxy using docker-compose:
docker-compose up --build

You will see the following error complaining:
```[2020-08-21 18:49:31.113][6][critical][main] [external/envoy/source/server/server.cc:98] error initializing configuration '/etc/envoy.yaml': Failed to load WASM module due to a missing import: __wbindgen_externref_xform__.__wbindgen_externref_table_grow```

### LINKED ISSUE #140: Runnable examples
This is more a discussion than an issue. Would it be valuable to move every example into it's own folder and include a matching envoy config + docker compose config? This should allow developers exploring the proxy-wasm SDK to run the example easily.

If this is considered valuable, I can contribute the changes.

### PR DESCRIPTION
**Pull request by Swaagie**
Migrate examples to separate folders.

**TODO:**

- [x] migrate hello_world
- [x] migrate http_auth_random
- [x] migrate http_headers
- [x] migrate http_body
- [x] migrate http_config
- [x] Build each example as part of CICD ~(and perhaps run?)~
- [x] `cargo audit` each `cargo.toml` for examples
- [x] Define documentation on how to build and run each example (per example or general?)

**Earlier draft PR discussion**

1. Single docker-compose that uses and environment variable to load the matching example. Upside: re-use of docker compose config. Downside: tiny bit of magic to newcomers and newer examples might require more elaborate docker compose config for them to make sense/work.
2. Do we want a separate `Cargo.toml` files for each example or go with Rust standard approach? It sounds like co-locating (dev-)dependencies was preferred to "avoid" the rustsec vuln from `chrono` from the main Crate? It would likely require more extensive changes and transition to a workspace.
3. To what extend does `raze` need to be involved? For devs coming from a mostly Rust background that could be tooling they are unfamiliar with.

fixes #140 
fixes #35 by showing how `dispatch_http_call` works

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -22,6 +22,20 @@
 - [HTTP Response body](./examples/http_body.rs)
 - [HTTP Configuration](./examples/http_config.rs)
 
+### Using WASM filters in Envoy
```
PiotrSikora, line 25: `s/WASM filters/Proxy-Wasm plugins`

PiotrSikora (commented):
> 1. Single docker-compose that uses and environment variable to load the matching example. Upside: re-use of docker compose config. Downside: tiny bit of magic to newcomers and newer examples might require more elaborate docker compose config for them to make sense/work.

See below (both answers). On one hand examples should be self-contained, on the other hand, Docker and Envoy might add unnecessary compexity as well...

> 2. Do we want a separate `Cargo.toml` files for each example or go with Rust standard approach? It sounds like co-locating (dev-)dependencies was preferred to "avoid" the rustsec vuln from `chrono` from the main Crate? It would likely require more extensive changes and transition to a workspace.

I'd prefer a separate `Cargo.toml` and `src/lib.rs` (since it has to be a `cdylib` for now), so that each example is a self-contained crate that users can copy as a base and evolve into their own plugins.

Working `cargo build --examples` would be great, but that's a secondary concern.

> 3. To what extend does `raze` need to be involved? For devs coming from a mostly Rust background that could be tooling they are unfamiliar with.

Agreed, Bazel integration adds a ton of complexity that isn't strictly necessary. Maybe let's ignore it for now in examples.

FWIW, we can probably land this after v0.2.0 (unless you want to clear `cargo-audit` warnings in dependent crates?).

Swaagie:
I think this would be fine to land after 0.2, just so we don't block the WASM reactor change going out. Will allow us to iterate a bit and land on a solid implementation we are all happy with. In addition, would enable you to bring in your uncommitted examples.

.github/workflows/rust.yml
```diff
@@ -250,4 +250,4 @@ jobs:
     - name: Run cargo audit
       run: |
         cp -p bazel/cargo/Cargo.raze.lock .
-        cargo audit --ignore RUSTSEC-2020-0159
```
Swaagie, line 253: Should the workflow be updated to also audit each example?

Cargo.toml
```diff
@@ -34,10 +27,10 @@ package_aliases_dir = "bazel/cargo"
 workspace_path = "//bazel/cargo"
 genmode = "Remote"
 
-[[example]]
-name = "hello_world"
-path = "examples/hello_world.rs"
-crate-type = ["cdylib"]
+# [[example]]
```
Swaagie, line 30: Unlikely this can stay due to example's `Cargo.toml` now having dependency definitions. Commented out first for clarity but should just be removed.

jcrugzz (approved):

PiotrSikora (commented):
Hi @Swaagie,
sorry for the delay, it looks that I forgot to send my review comments.

Also, could you incorporate `strip` option from https://github.com/proxy-wasm/proxy-wasm-rust-sdk/pull/160?

Thanks!

PiotrSikora (approved):
Thanks!

PR #151, opened 2022-04-30
chore: enable dep bot

haoxins:


PiotrSikora:
Hi @haoxins! Thanks for the contribution.

However:
- There is already a daily `cargo outdated` check (scheduled via GitHub Actions) that acts as a notification for the outdated dependencies.
- AFAIK, dependabot is not Bazel-aware, so any PRs would need to be closed and created manually anyway.

As such, I don't see any value in enabling this.

haoxins:
glad to know that ~

PR #152, opened 2022-04-30
chore: uses edition 2021

haoxins:


PiotrSikora (commented):
This is fine, since `hashbrown` dependency already forces us to use a Rust compiler that supports `edition = 2021`, but I'm wondering if `edition` change requires a version bump (preferrably not)?

Also, there are a few more places where `edition` needs to be updated for Bazel builds (`BUILD` and `examples/BUILD`). 

cc @Swaagie who suggested this change earlier in #129 and #133, but we dropped it since it wasn't required after all.

Swaagie:
Probably good to move forward with latest edition. Test and builds were passing with 2021, but as @PiotrSikora mentioned it requires a few more updates.

PR #154, opened 2022-05-23
Don't panic on unknown token_id in on_grpc_close.

### LINKED ISSUE #153: `hostcalls::open_grpc_stream` does not register closed streams, upstream failures result in panic
At the moment:
* `hostcalls::open_grpc_stream` only registers stream IDs when connections are successful
* Envoy calls close_grpc_stream on the stream id that is passed in, even when the original `open_grpc_stream` failed.
* `close_grpc_stream` can't route to the proper context and panics on `invalid token_id`

This seems to be an issue of ambiguous specification. I will open a PR to fix this, but it may be considered that Envoy is in the wrong here instead. Either way, we probably shouldn't panic.

Update:
* Envoy doesn't return the token ID, but still fires the on_close event. We will just need to ignore/log in on_grpc_close.

Protryon:
Fixes #153. 

This PR:
* Logs in `on_grpc_close` when an invalid `token_id` is sent from the host. This happens in Envoy when the initial connection fails.

### DIFF
--- a/src/dispatcher.rs
+++ b/src/dispatcher.rs
@@ -16,7 +16,7 @@ use crate::hostcalls;
 use crate::traits::*;
 use crate::types::*;
 use hashbrown::HashMap;
-use log::warn;
+use log::trace;
 use std::cell::{Cell, RefCell};
 
 thread_local! {
@@ -536,7 +536,8 @@ impl Dispatcher {
                 root.on_grpc_stream_close(token_id, status_code)
             }
         } else {
-            warn!("on_grpc_close: invalid token_id, a non-connected stream has closed");
+            //TODO: change back to a panic once underlying issue is fixed.
+            trace!("on_grpc_close: invalid token_id, a non-connected stream has closed");
         }
     }
 }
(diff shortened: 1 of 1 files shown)


Protryon:
@PiotrSikora Any updates on this?

src/dispatcher.rs
```diff
@@ -535,7 +536,7 @@ impl Dispatcher {
                 root.on_grpc_stream_close(token_id, status_code)
             }
         } else {
-            panic!("invalid token_id")
+            warn!("on_grpc_close: invalid token_id, a non-connected stream has closed");
```
PiotrSikora, line 539: Since this is a known behavior, I'd lower this to the `trace` level.

Could you also add a TODO mentioning to restore the `panic` (maybe just comment it out) once the issue is fixed in Envoy?

Lastly, I'd really appreciate if you could open on issue with Envoy as well, since it should be fixed there.

fmcato:
We might be hit by this when using unary grpc calls (seems to be triggered by invalid calls due to bad GRPC upstream configuration).

Let me know if I can help here.

src/dispatcher.rs
Protryon: @PiotrSikora Done. Changed log level to trace and added a TODO.

PiotrSikora (approved):
Could you merge with `master`? Thanks!

PR #157, opened 2022-07-05
Add missing parameter to on_http_response_trailers

casimiro:
Proxy wasm spec defines `proxy_on_http_trailers` as a 3 parameter function: `context_id, num_trailers, end_of_stream`.
This PR adds the missing `end_of_stream` parameter.


PiotrSikora:
Hi @casimiro, apologies for the late response, and thanks for the contribution!

The current code is fine (otherwise, it wouldn't work with the existing host implementations). The issue is that there is no written specification for the Proxy-Wasm ABI v0.2.1, which is implemented by Envoy, MSON and all the SDKs (but writting it down is my primary task for the next 2 weeks).

Where did you run into this issue?

casimiro:
Hi @PiotrSikora, I apologize for the ultra late response! I ran into this when experimenting with adding http trailers support to ngx_wasm_module. I assumed the `end_of_stream` argument was missing based on the late vNEXT spec which included `end_of_stream` as part of `proxy_on_http_response_trailers`.

Thank you for the response.

PR #160, opened 2022-08-01
Strip "debuginfo" by default.

### LINKED ISSUE #149: debuginfo is not stripped when using `bazel build -c opt`
build the hello_world example in the project:
```bash
bazel build -c opt //examples:hello_world
```

use `wasm2wat` to generate the wat file:
```bash
wasm2wat bazel-bin/examples/hello_world.wasm -o hello_world.wat
```

then you will find that the debuginfos is not stripped:
```text
...
  (import "wasi_snapshot_preview1" "environ_sizes_get" (func $__imported_wasi_snapshot_preview1_environ_sizes_get (type 4)))
  (import "wasi_snapshot_preview1" "proc_exit" (func $__imported_wasi_snapshot_preview1_proc_exit (type 2)))
  (func $__wasm_call_ctors (type 18)
    call $__wasilibc_initialize_environ_eagerly)
  (func $_ZN4core3ops8function6FnOnce9call_once17h2428394f5d7238e6E (type 3) (param i32 i32)
    local.get 0
    i32.const 1048980
    i32.store offset=4
    local.get 0
    i32.const 1
    i32.store)
  (func $_ZN4core3ptr44drop_in_place$LT$getrandom..error..Error$GT$17h161e8c8ed28efdccE (type 2) (param i32))
  (func $_ZN75_$LT$hello_world..HelloWorld$u20$as$u20$proxy_wasm..traits..RootContext$GT$11on_vm_start17h33e5123357b23fd4E (type 4) (param i32 i32) (result i32)
...
```

It increases the size of the wasm file and reduces performance. The `hello_world.wasm` is about 1.9Mb:
image

I use this project ([hello-world-wasm.zip](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/files/8551257/hello-world-wasm.zip)) and build the wasm file with `wasm-pack build --release`, the wasm file size is only 80Kb:

image

PiotrSikora:
Fixes #149.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -100,32 +100,32 @@ jobs:
 
     - name: Build (wasm32-unknown-unknown)
       env:
-        RUSTFLAGS: -D warnings -C link-args=-S
+        RUSTFLAGS: -D warnings
       run: cargo build --release --all-targets --target=wasm32-unknown-unknown
 
     - name: Clippy (wasm32-unknown-unknown)
       env:
-        RUSTFLAGS: -D warnings -C link-args=-S
+        RUSTFLAGS: -D warnings
       run: cargo clippy --release --all-targets --target=wasm32-unknown-unknown
 
     - name: Build (wasm32-wasi)
       env:
-        RUSTFLAGS: -D warnings -C link-args=-S
+        RUSTFLAGS: -D warnings
       run: cargo build --release --all-targets --target=wasm32-wasi
 
     - name: Build (wasm32-wasi with wee-alloc)
       env:
-        RUSTFLAGS: -D warnings -C link-args=-S
+        RUSTFLAGS: -D warnings
       run: cargo build --release --all-targets --target=wasm32-wasi --features=wee-alloc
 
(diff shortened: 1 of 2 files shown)


PiotrSikora:
cc @johnlanni

.github/workflows/rust.yml
```diff
@@ -100,32 +100,32 @@ jobs:
 
     - name: Build (wasm32-unknown-unknown)
       env:
-        RUSTFLAGS: -D warnings -C link-args=-S
```
mpwarres, line 103: To confirm, the reason `-S` is removed from linker arguments is that we are now relying on debug symbol stripping by rustc, and also linker `-S` was ineffective in removing debug symbols?

mpwarres (approved):

.github/workflows/rust.yml
PiotrSikora: `-C link-args=-S` was effective, but it was running only as part of the CI.

mpwarres: Got it, thanks for the explanation.

PR #176, opened 2022-11-18
Update rules_rust to v0.5.0 (with Rust v1.61.0).

PiotrSikora:
Signed-off-by: Piotr Sikora <piotr@aviatrix.com>

### DIFF
--- a/bazel/repositories.bzl
+++ b/bazel/repositories.bzl
@@ -19,7 +19,8 @@ def proxy_wasm_rust_sdk_repositories():
     maybe(
         http_archive,
         name = "rules_rust",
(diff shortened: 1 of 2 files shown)


mpwarres (approved):
LGTM, though looks like a couple of the presubmit checks failed?

PiotrSikora:
> LGTM, though looks like a couple of the presubmit checks failed?

cargo-raze v0.15.0 doesn't work with recent versions or rules_rust, and
cargo-raze v0.16.0 generates broken BUILD rules, so we cannot update it.

rules_rust v0.5 ships with Rust v1.61.0 (which is going to be our MSRV)
and works with cargo-raze v0.15.0, so let's use that for the time being.

PR #190, opened 2023-03-05
try if trust-dns-resolver works or not

Cindia-blue:


examples/http_headers/src/lib.rs
```diff
@@ -43,10 +74,25 @@ impl Context for HttpHeaders {}
...
+        
+        let lookup = LookupFuture::lookup(
+            vec![Name::from_str("www.example.com.").unwrap()],
+            RecordType::A,
+            Default::default(),
+            CachingClient::new(0, client, false),
+        );
+    
+        let io_loop = Runtime::new().unwrap();
+        let lookup = io_loop.block_on(lookup).unwrap();
```
PiotrSikora, line 86: This seems like a bad idea, since each worker thread (in case of Envoy) is going to be blocked on this, and DNS resolution can potentially take hundreds of milliseconds... not to mention the overhead of starting Tokio runtime per each request.

PiotrSikora (changes requested):
I'm generally fine with adding examples like that, but could you create a new example instead of overloading the existing one with unrelated test code?

Also, the outgoing DNS request won't be sent, since Proxy-Wasm doesn't yet allow TCP/UDP sidecalls. Did you try to use this code in Envoy or another host implementation?

examples/http_headers/src/lib.rs
Cindia-blue: Can we have a meeting about this? Can we mitigate the overhead per request by cache layer?

PiotrSikora: You could store the runtime in a global variable and share it across the callbacks, but I suspect that Tokio runtime couldn't make any progress without deep integration with Proxy-Wasm's callbacks.

Also, if you try to compile your example, you can see that Tokio async doesn't work in Wasm (yet):
```
compile_error!("Only features sync,macros,io-util,rt,time are supported on wasm.");
```
What are you trying to achieve? What's your end goal?

Cindia-blue: yup... I can  not compile through for that reason. I am trying to get cname chain in WASM. Could you recommend some feasible way? I saw there are some other lib besides Tokio, RustDNS but not sure if it has similar issue. It will be nice if you can suggest

PiotrSikora: Are you only interested in the final answer or the whole CNAME chain? If the former, and you're using Envoy, then there are `envoy_resolve_dns` / `envoy_on_resolve_dns` extensions that you could call and use Envoy's built-in DNS resolver... Note that they are not exposed in the Rust SDK yet.

Issue #206, opened 2023-08-21
Panic when there are no remote hosts


erikness-doordash:
We get a panic in our Rust Envoy filter when we try to open a new gRPC stream to any hostname that DNS can’t find any hosts for. Reproduced with Envoy versions 1.25.2 and 1.28.2, and the latest Rust proxy-wasm version 0.2.1.

Panic happens when:
1. we point the cluster config at a fake hostname (dns domain is that not resolvable)
2. we point the cluster config at a DNS hostname that returns 0 healthy hosts (e.g. service scaled down to zero)

There is NOT a panic if there is a server found and returns gRPC UNAVAILABLE.

Detailed order of events:

1. We call [Context.open_grpc_stream()](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/6b47aec926bc29971c727471d6f4c972ec407c7f/src/traits.rs#L146C8-L146C24) from this library
2. Proxy WASM calls [proxy_grpc_stream()](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/6b47aec926bc29971c727471d6f4c972ec407c7f/src/hostcalls.rs#L854) on the Envoy host
3. Envoy host receives gRPC unavailable
4. Proxy WASM receives Status::InternalFailure ([host side code](https://github.com/envoyproxy/envoy/blob/afa98867c807dee0d833da701ba3ab0c9ace9ada/source/extensions/common/wasm/context.cc#L1042) / [proxy-wasm-rust-sdk code](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/6b47aec926bc29971c727471d6f4c972ec407c7f/src/hostcalls.rs#L870))
    1. As a result the stream’s token ID is not saved to the Proxy WASM `Dispatcher.grpc_streams` HashMap
5. Proxy WASM gets a call from host to its method `Dispatcher.on_grpc_receive_initial_metadata()` with token number of the stream that couldn’t get created
6. Token ID not in hashmap, PANIC

erikness-doordash:
**PANIC SCENARIO LOGS**

```
starting main dispatch loop
async DNS resolution complete for [HOSTNAME]
DNS refresh rate reset for [HOSTNAME], (failure) refresh rate 5000 ms
initializing Primary cluster [SERVICE_NAME] completed
init manager Cluster [SERVICE_NAME] contains no targets
[...]

PR #210, opened 2023-10-18
Don't panic on unknown token_id in gRPC callbacks.

erikness-doordash:
Sometimes `Dispatcher.on_grpc_receive_initial_metadata()` will get an invalid stream ID, and then panic. `Dispatcher.on_grpc_receive_trailing_metadata()` as well.

This is a quick fix that will turn a `panic!()` call into a `trace!()` call. The broader issue/bug is out of scope here, though as far as I know it's benign other than this panic.

~NOTE: there may be a more concise way to write this Rust code.~ (thanks Piotr)
NOTE: adding this change to `on_grpc_receive()` as well, just to be safe, even though I haven't seen this panic get triggered myself.

See:
1. Issue with more details: https://github.com/proxy-wasm/proxy-wasm-rust-sdk/issues/206
2. Issue tracking why these two methods get an invalid token number at all: https://github.com/proxy-wasm/proxy-wasm-cpp-host/issues/373

### DIFF
--- a/src/dispatcher.rs
+++ b/src/dispatcher.rs
@@ -491,11 +491,17 @@ impl Dispatcher {
     }
 
     fn on_grpc_receive_trailing_metadata(&self, token_id: u32, trailers: u32) {
-        let context_id = *self
-            .grpc_streams
-            .borrow_mut()
-            .get(&token_id)
-            .expect("invalid token_id");
+        let grpc_streams_ref = self.grpc_streams.borrow_mut();
+        let context_id_hash_slot = grpc_streams_ref
+            .get(&token_id);
+        let context_id = match context_id_hash_slot {
+            Some(id) => *id,
+            None => {
+                // TODO: change back to a panic once underlying issue is fixed.
+                trace!("on_grpc_receive_initial_metadata: invalid token_id");
+                return;
+            }
+        };
 
         if let Some(http_stream) = self.http_streams.borrow_mut().get_mut(&context_id) {
             self.active_id.set(context_id);
--- a/src/dispatcher.rs
+++ b/src/dispatcher.rs
@@ -429,8 +429,7 @@ impl Dispatcher {
 
     fn on_grpc_receive_initial_metadata(&self, token_id: u32, headers: u32) {
         let grpc_streams_ref = self.grpc_streams.borrow_mut();
-        let context_id_hash_slot = grpc_streams_ref
-            .get(&token_id);
+        let context_id_hash_slot = grpc_streams_ref.get(&token_id);
         let context_id = match context_id_hash_slot {
             Some(id) => *id,
             None => {
@@ -492,8 +491,7 @@ impl Dispatcher {
 
     fn on_grpc_receive_trailing_metadata(&self, token_id: u32, trailers: u32) {
         let grpc_streams_ref = self.grpc_streams.borrow_mut();
-        let context_id_hash_slot = grpc_streams_ref
-            .get(&token_id);
+        let context_id_hash_slot = grpc_streams_ref.get(&token_id);
         let context_id = match context_id_hash_slot {
             Some(id) => *id,
             None => {
--- a/src/dispatcher.rs
+++ b/src/dispatcher.rs
@@ -428,11 +428,17 @@ impl Dispatcher {
     }
 
     fn on_grpc_receive_initial_metadata(&self, token_id: u32, headers: u32) {
-        let context_id = *self
-            .grpc_streams
-            .borrow_mut()
-            .get(&token_id)
-            .expect("invalid token_id");
+        let grpc_streams_ref = self.grpc_streams.borrow_mut();
+        let context_id_hash_slot = grpc_streams_ref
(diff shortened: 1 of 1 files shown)


src/dispatcher.rs
```diff
@@ -428,11 +428,17 @@ impl Dispatcher {
...
-            .borrow_mut()
-            .get(&token_id)
-            .expect("invalid token_id");
+        let grpc_streams_ref = self.grpc_streams.borrow_mut();
+        let context_id_hash_slot = grpc_streams_ref
+            .get(&token_id);
+        let context_id = match context_id_hash_slot {
+            Some(id) => *id,
+            None => {
+                // TODO: change back to a panic once underlying issue is fixed.
```
erikness-doordash, line 437: This matches the language in a very similar PR a while back: https://github.com/proxy-wasm/proxy-wasm-rust-sdk/pull/154

src/dispatcher.rs
```diff
@@ -428,11 +428,17 @@ impl Dispatcher {
...
+                trace!("on_grpc_receive_initial_metadata: invalid token_id");
```
erikness-doordash, line 438: do we want to print the value of the token id? would it be useful?

erikness-doordash:
Just tested this, no more crash after scaling a remote service to 0

src/dispatcher.rs
PiotrSikora: I don't think so.

src/dispatcher.rs
```diff
@@ -485,11 +490,16 @@ impl Dispatcher {
     }
```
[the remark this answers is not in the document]
PiotrSikora: Actually, maybe just add it anyway just to stay on the safe side. Thanks!

PiotrSikora:
Could you fix DCO?

erikness-doordash:
## Code search to support patching `on_grpc_receive()` as well

I went looking through the host code; I couldn't find where exactly Envoy calls these methods in the unhappy path (can't init stream), it may be pretty low level.

But looking at this call stack, I *suspect* `on_grpc_receive()` needs this too:

1. (proxy-wasm-cpp-host) ContextBase::onGrpcReceiveInitialMetadata
2. (envoy) Context::onGrpcReceiveInitialMetadataWrapper()
3. (envoy) GrpcStreamClientHandler::onReceiveInitialMetadata()
    1. AsyncStreamImpl::onHeaders()
        1. AsyncStreamImpl::encodeHeaders()
        2. ???
    2. GoogleAsyncStreamImpl::handleOpCompletion()
        - **This has a branch for 2 methods on the "callbacks_" object: `onReceiveInitialMetadata()` and `onReceiveMessageRaw()`**

(similar for trailing metadata)

Though I have not seen `on_grpc_receive()` calls triggered in our own code in this circumstance (see: https://github.com/proxy-wasm/proxy-wasm-rust-sdk/issues/206).

src/dispatcher.rs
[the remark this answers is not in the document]
erikness-doordash: Agreed, I put some supporting evidence in this comment: https://github.com/proxy-wasm/proxy-wasm-rust-sdk/pull/210#issuecomment-1779881518

src/dispatcher.rs
```diff
@@ -485,11 +490,16 @@ impl Dispatcher {
...
 
     fn on_grpc_receive_trailing_metadata(&self, token_id: u32, trailers: u32) {
-        let context_id = *self
-            .grpc_streams
-            .borrow_mut()
-            .get(&token_id)
-            .expect("invalid token_id");
+        let grpc_streams_ref = self.grpc_streams.borrow_mut();
+        let context_id_hash_slot = grpc_streams_ref.get(&token_id);
+        let context_id = match context_id_hash_slot {
```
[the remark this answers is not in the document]
erikness-doordash: For some reason I thought the rust compiler would stop me. Checked again though, it works fine, thanks for the tip

src/dispatcher.rs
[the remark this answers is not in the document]
erikness-doordash: Done

src/dispatcher.rs
```diff
@@ -480,16 +483,21 @@ impl Dispatcher {
                 root.on_grpc_stream_message(token_id, response_size);
             }
         } else {
-            panic!("invalid token_id")
+            // TODO: change back to a panic once underlying issue is fixed.
+            trace!("on_grpc_receive_initial_metadata: invalid token_id");
+            return;
```
PiotrSikora, line 488: Nit: the `return` here isn't needed.

PiotrSikora (commented):
LGTM, sans the nit.

src/dispatcher.rs
erikness-doordash: Oh yeah, thanks for catching, fixed

erikness-doordash:
> LGTM, sans the nit.

@PiotrSikora can you merge it?

PiotrSikora (approved):
Thanks!

Issue #148, opened 2022-04-14
Setting gRPC status in http response


wagenet:
Is this possible? It looks like the underlying interface may support it, but it doesn’t seem to be public.

juanmolle:
Do we need to expose? maybe instead of return -1 https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/master/src/hostcalls.rs#L721 we can map http code to the closest grpc code

juanmolle:
@PiotrSikora let us know what you think about this please

PiotrSikora:
This should be fixed in #248.

Note that Envoy doesn't support sending gRPC messages using `sendLocalReply()` calls, and it automatically changes HTTP responses to gRPC errors when talking with gRPC clients, so this is limited to emitting gRPC error responses, and not for producing valid gRPC responses.

PR #211, opened 2023-10-24
Adding http to grpc status code mapping

juanmolle:
Instead of sending unrelated -1 grps status code map the http code into a representative grpc code.

Related issue @https://github.com/proxy-wasm/proxy-wasm-rust-sdk/issues/148

### DIFF
--- a/src/hostcalls.rs
+++ b/src/hostcalls.rs
@@ -703,27 +703,6 @@ extern "C" {
     ) -> Status;
 }
 
-#[allow(dead_code)]
-enum GrpcStatucCode {
-    Ok = 0,
-    Cancelled = 1,
-    Unknown = 2,
-    InvalidArgument= 3,
-    DeadlineExceeded = 4,
-    NotFound = 5,
-    AlreadyExists = 6,
-    PermissionDenied = 7,
-    ResourceExhausted = 8,
-    FailedPrecondition = 9,
-    Aborted = 10,
-    OutOfRange = 11,
-    Uninmplemented = 12,
-    Internal = 13,
-    Unavailable = 14,
-    DataLoss = 15,
-    Unauthenticated = 16,
-}
-
 fn http_to_grpc_status_code(status_code: u32) -> i32 {
     match status_code {
         200 => GrpcStatucCode::Ok as i32,
--- a/src/hostcalls.rs
+++ b/src/hostcalls.rs
@@ -703,6 +703,44 @@ extern "C" {
     ) -> Status;
 }
 
+#[allow(dead_code)]
+enum GrpcStatucCode {
+    Ok = 0,
+    Cancelled = 1,
+    Unknown = 2,
+    InvalidArgument= 3,
+    DeadlineExceeded = 4,
+    NotFound = 5,
+    AlreadyExists = 6,
+    PermissionDenied = 7,
+    ResourceExhausted = 8,
+    FailedPrecondition = 9,
+    Aborted = 10,
+    OutOfRange = 11,
+    Uninmplemented = 12,
+    Internal = 13,
+    Unavailable = 14,
+    DataLoss = 15,
+    Unauthenticated = 16,
+}
+
+fn http_to_grpc_status_code(status_code: u32) -> i32 {
+    match status_code {
+        200 => GrpcStatucCode::Ok as i32,
+        400 => GrpcStatucCode::InvalidArgument as i32,
+        401 => GrpcStatucCode::Unauthenticated as i32,
+        403 => GrpcStatucCode::PermissionDenied as i32,
+        404 => GrpcStatucCode::NotFound as i32,
+        409 => GrpcStatucCode::Aborted as i32,
+        412 => GrpcStatucCode::FailedPrecondition as i32,
+        429 => GrpcStatucCode::ResourceExhausted as i32,
+        500 => GrpcStatucCode::Internal as i32,
+        501 => GrpcStatucCode::Uninmplemented as i32,
+        503 => GrpcStatucCode::Unavailable as i32,
+        _ => GrpcStatucCode::Unknown as i32,
+    }
+}
+
 pub fn send_http_response(
     status_code: u32,
     headers: Vec<(&str, &str)>,
@@ -718,7 +756,7 @@ pub fn send_http_response(
             body.map_or(0, |body| body.len()),
             serialized_headers.as_ptr(),
             serialized_headers.len(),
-            -1,
+            http_to_grpc_status_code(status_code),
         ) {
             Status::Ok => Ok(()),
             status => panic!("unexpected status: {}", status as u32),
--- a/src/types.rs
+++ b/src/types.rs
@@ -115,3 +115,24 @@ pub enum MetricType {
 }
 
 pub type Bytes = Vec<u8>;
+
+#[allow(dead_code)]
+enum GrpcStatucCode {
+    Ok = 0,
+    Cancelled = 1,
+    Unknown = 2,
+    InvalidArgument = 3,
+    DeadlineExceeded = 4,
+    NotFound = 5,
+    AlreadyExists = 6,
+    PermissionDenied = 7,
+    ResourceExhausted = 8,
+    FailedPrecondition = 9,
+    Aborted = 10,
+    OutOfRange = 11,
+    Uninmplemented = 12,
+    Internal = 13,
+    Unavailable = 14,
+    DataLoss = 15,
+    Unauthenticated = 16,
+}


src/hostcalls.rs
```diff
@@ -703,6 +703,44 @@ extern "C" {
     ) -> Status;
 }
 
+#[allow(dead_code)]
+enum GrpcStatucCode {
```
juanmolle, line 707: should this enum be moved to types?

src/hostcalls.rs
```diff
@@ -718,7 +756,7 @@ pub fn send_http_response(
             body.map_or(0, |body| body.len()),
             serialized_headers.as_ptr(),
             serialized_headers.len(),
-            -1,
+            http_to_grpc_status_code(status_code),
```
PiotrSikora, line 759: I don't believe this is correct. It appear to be always sending gRPC status code, even for non-gRPC responses.

juanmolle: that's true, is it possible to know if it is grpc in this point? or in the wasm calling this? If I create a wasm that just create a local reply and I make a call with grpcurl, the error I get is the code -1, then current implementation is always sending an status code and envoy is forwarding to the client. How do you think we should fix this scenario?

juanmolle: Interesting, there is already a mapping in envoy (not the same I have created) 
https://github.com/envoyproxy/envoy/blob/main/source/common/grpc/status.cc#L6

and I guess from envoy doc,  will use this value to override that
https://github.com/envoyproxy/envoy/blob/main/envoy/http/filter.h#L579

What seams to be happening is that in the wasm context the value sent from this function (-1). is directly sent here
https://github.com/envoyproxy/envoy/blob/main/source/extensions/common/wasm/context.cc#L1672

and set always a value to the optional, and does not let envoy to automatically map to the corresponding grpc code from http code, because any value sent will override that

juanmolle: probably it would be cleaner if we can instruct envoy to interpret -1 as null (because 0 is a valid status code) to forward that an an empty optional and let envoy do it's magic., saying that, I still think that sending a meaning value could be better than sending -1.

juanmolle: Let me shared the test I have done, first a call without the change, and after that calls with a wasm filter that do just basic auth, the first request show the -1 (4294967295) currently sent by the wasm without the change

Old wasm

jolle@jolle-ltmrmbv InstallFromScratch % grpcurl -d '{"greeting": "pepe"}' --plaintext -protoset grpc/hello.proto-descriptor_grpc localhost:8080  hello.HelloService.SayHello
ERROR:
  Code: Internal
  Message: transport: malformed grpc-status: strconv.ParseInt: parsing "4294967295": value out of range

New Wasm rejecting the call no auth header

% grpcurl -d '{"greeting": "pepe"}' --plaintext -protoset grpc/hello.proto-descriptor_grpc localhost:8080  hello.HelloService.SayHello
ERROR:
  Code: Unauthenticated
  Message: {"error":"Registered authentication is set to HTTP basic authentication but there was no security context on the session."}

New wasm with invalid user:pass

% grpcurl -expand-headers -H 'Authorization: Basic YXNkOmFzZA==' -d '{"greeting": "mulesoft"}' --plaintext -protoset grpc/hello.proto-descriptor_grpc localhost:8080  hello.HelloService.SayHello
ERROR:
  Code: Unauthenticated
  Message: {"error":"Authentication Attempt Failed"}


Server rejected the query (proto not corresponding with server)

% grpcurl -expand-headers -H 'Authorization: Basic YWRtaW46dGVzdA==' -d '{"greeting": "pepe"}' --plaintext -protoset grpc/hello.proto-descriptor_grpc localhost:8080  hello.HelloService.SayHello
ERROR:
  Code: Unimplemented
  Message: unknown service hello.HelloService

OK

% grpcurl -expand-headers -H 'Authorization: Basic YWRtaW46dGVzdA==' -d '{"greeting": "pepe"}' --plaintext -protoset grpc/hello.proto-descriptor_grpc localhost:8080  hello.HelloService.SayHello
{
  "reply": "hello pepe"
}

juanmolle:
@PiotrSikora any thought about the explanation and use cases I have described?

PiotrSikora:
> @PiotrSikora any thought about the explanation and use cases I have described?

This seems to be a real issue that needs to be fixed, but the solution is not to emit any `grpc-status` code when not requested.

Emitting `grpc-status` code when sending a plain HTTP (i.e. non-gRPC) response is trading one incorrect behavior for another.

I suspect there was a regression on the Envoy side, since I cannot imagine we were emitting `grcp-status = -1` from the beginning. @mpwarres could you please take a look?

PiotrSikora:
Thinking about this a bit more, maybe adding `send_grpc_response` where users can provide `grpc-status` would be a better solution?

juanmolle:
not sure adding a new method, many samples does not know in the wasm if the request comes from a grpc client. for example native ext_authz does not include a grpc status code but works with grpcurl. due to the fact that envoy does not support optionals, the other option is changing the signature, but unfortunately will change the interface and I guess there are no default values nor overload like there are in c++. I still thinking for viable options

PiotrSikora:
> not sure adding a new method, many samples does not know in the wasm if the request comes from a grpc client.

I guess the question is what problem are you trying to solve?

Based on the PR, I assumed you wanted to emit custom gRPC responses with correct `grpc-status`.

However, your response suggests that you want to use `grpcurl` tool to work with plain HTTP (i.e. non-gRPC) responses? While emitting `grpc-status` isn't great, it's a "legitimate" HTTP header, so it shouldn't be breaking anything at the HTTP level.

Also, per [gRPC spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md), the only valid HTTP response code is `200 OK`, so the `grpc-status` for non-200 status codes shouldn't be interpreted anyway.

juanmolle:
>  guess the question is what problem are you trying to solve?

When a grpc client try to connect to grpc server and a wasm filter is in the filter chain that replay with unautorized, the wasm replay with -1 that is an unknown grpc status code, no letting the clien know what had happened

'''
Code: Internal
Message: transport: malformed grpc-status: strconv.ParseInt: parsing "4294967295": value out of range
'''

This is not happening with native filters like ext_authz, because the filter sent null as optional value and envoy translate the http unauthorized code into the equivalent grpc status

But I get your point. I will go to envoy again to in witch condition the grpc status code is infer from the http status code.

just to be in the same page the correct fix should be in envoy wasm filter to allows the grpc code be optional, but that request an api change that I guess will to be easy to do. In addition forcing the grpc status code to be always -1 is incorrect too and I though a situation in the middle when instead for sending invalid -1 was better.

The other not elegant workaround is to translate the -1 into empty optional in the wasm filter, but the multiple conversions between int32 -> uint32 -> uint64  make even less elegant. ( and I don't really know if all arch will behave equal)

PiotrSikora:
> When a grpc client try to connect to grpc server and a wasm filter is in the filter chain that replay with unautorized, the wasm replay with -1 that is an unknown grpc status code, no letting the clien know what had happened
> 
> ''' Code: Internal Message: transport: malformed grpc-status: strconv.ParseInt: parsing "4294967295": value out of range '''

Right, but the correct response in that case is `200 OK` with `grpc-status: UNAUTHENTICATED` or `grpc-status: PERMISSION_DENIED`, not `401 Unauthorized`.

And since the existing ABI doesn't support trailers in HTTP callouts, the only way to emit that from the Rust SDK would be to add `send_grpc_response`, as mentioned earlier.

> This is not happening with native filters like ext_authz, because the filter sent null as optional value and envoy translate the http unauthorized code into the equivalent grpc status
> 
> But I get your point. I will go to envoy again to in witch condition the grpc status code is infer from the http status code.
>
> just to be in the same page the correct fix should be in envoy wasm filter to allows the grpc code be optional, but that request an api change that I guess will to be easy to do. In addition forcing the grpc status code to be always -1 is incorrect too and I though a situation in the middle when instead for sending invalid -1 was better.
> 
> The other not elegant workaround is to translate the -1 into empty optional in the wasm filter, but the multiple conversions between int32 -> uint32 -> uint64 make even less elegant. ( and I don't really know if all arch will behave equal)

That's an issue in Envoy integration, and not Rust SDK. Also, I don't believe it requires any API changes, since Envoy's host integration can simply map `-1` (which is invalid value) to `std::nullopt`. cc @mpwarres

FWIW, the mapping is not a that bad, since we're sending garbage HTTP header already, but if you want to send gRPC responses, then you need to add `send_grpc_response` that allows users to set gRPC status.

PiotrSikora:
> It makes sense to me and let envoy replay automatically, If you agree I could make envoy change to do it.

Please do, thank you!

PR #212, opened 2023-10-26
Add note about installing required targets to README

petedmarsh:
These are a pre-requisit for builds to complete.


petedmarsh:
This tripped me up a little, wether that is mostly a me problem or this is something that would be helpful for others I can't say, but I thought I'd raise this just in case :)

PR #216, opened 2023-11-03
The current code needs to allocate a new Vec when a property is accessed

andresmargalef:
This patch add a new struct PropertyPath and a macro used to create a PropertyPath with a slice of &str using const generics to prevent allocations in runtime.

Before:
```rust
let property_path = vec!["path", "to", "property"];
let property_bytes = match proxy_wasm::hostcalls::get_property(property_path) {
    Ok(value) => value,
    _ => return Action::Continue,
};

```

New method:
```rust
let property_path = create_property_path!("path", "to", "property");
let property_bytes = match proxy_wasm::hostcalls::get_property_by_path(&property_path) {
    Ok(value) => value,
    _ => return Action::Continue,
};
```

### DIFF
--- a/src/hostcalls.rs
+++ b/src/hostcalls.rs
@@ -17,6 +17,68 @@ use crate::types::*;
 use std::ptr::{null, null_mut};
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
+pub struct PropertyPath<const S: usize> {
+    path: [u8; S],
+}
+
+pub const fn serialized_property_path_len(path_segments: &[&str]) -> usize {
+    let mut sum = 0;
+    let mut i = 0;
+    while i < path_segments.len() {
+        sum += path_segments[i].len();
+        i += 1;
+    }
+
+    // add separator 0x0 between segments
+    if path_segments.len() > 1 {
+        sum += path_segments.len() - 1
+    }
+
+    sum
+}
+
+pub const fn build_serialized_property_path<const S: usize>(
+    path_segments: &[&str],
+) -> PropertyPath<S> {
+    let mut output = [0; S];
+    let mut output_idx = 0;
+    let mut path_segment_idx = 0;
+
+    while path_segment_idx < path_segments.len() {
+        let segment = path_segments[path_segment_idx].as_bytes();
+
+        let mut segment_idx = 0;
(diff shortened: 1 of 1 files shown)


PiotrSikora:
@andresmargalef this looks great, but I'm wondering what's the motivation behind this change? Does this result in measurable performance changes in your plugins?

andresmargalef:
The motivation is latency reduction removing extra work like allocations, right now i'm building an example application using wasmtime to test locally ours filters using this patch vs master. I will share any results here :D

andresmargalef:
The latency reduction is ~30%, the "toy" benchmark runs 1000 iterations and calculate the avg in nanos. The test call only one get_property using a vec vs get_property_path, and on the host side always returns 1 (NotFound).
I need to improve de code and use criterion to get better numbers.

```
request_metrics.wasm.optimized[1000 iterations]: 452 avg nanos
request_metrics.wasm.slow[1000 iterations]: 510 avg nanos
request_metrics.wasm.optimized[1000 iterations]: 278 avg nanos
request_metrics.wasm.slow[1000 iterations]: 634 avg nanos
request_metrics.wasm.optimized[1000 iterations]: 230 avg nanos
request_metrics.wasm.slow[1000 iterations]: 426 avg nanos
request_metrics.wasm.optimized[1000 iterations]: 196 avg nanos
request_metrics.wasm.slow[1000 iterations]: 278 avg nanos
request_metrics.wasm.optimized[1000 iterations]: 191 avg nanos
request_metrics.wasm.slow[1000 iterations]: 289 avg nanos
request_metrics.wasm.optimized[1000 iterations]: 191 avg nanos
request_metrics.wasm.slow[1000 iterations]: 276 avg nanos
request_metrics.wasm.optimized[1000 iterations]: 188 avg nanos
request_metrics.wasm.slow[1000 iterations]: 276 avg nanos
request_metrics.wasm.optimized[1000 iterations]: 188 avg nanos
request_metrics.wasm.slow[1000 iterations]: 277 avg nanos
```

andresmargalef:
Hi @PiotrSikora , do you think it makes sense to merge this pull request? Do not hesitate to tell me about any changes you see that need to be made as well as any other perf test.

PR #221, opened 2024-02-13
Update rules_rust to v0.38.0 (with Rust v1.75.0).

mikegrass:
Our rust wasm extensions builds are currently blocked with a dependency failure because the latest `ahash` wants rustc 1.72 or later:

```
11:54:16  error: package `ahash v0.8.8` cannot be built because it requires rustc 1.72.0 or newer, while the currently active rustc version is 1.67.0
11:54:16  Either upgrade to rustc 1.72.0 or newer, or use
11:54:16  cargo update -p ahash@0.8.8 --precise ver
11:54:16  where `ver` is the latest version of `ahash` supporting rustc 1.67.0
```

Pulling in rules_rust-v0.38.0 (rustc 1.75) and updating deps.


PiotrSikora:
> Our rust wasm extensions builds are currently blocked with a dependency failure because the latest `ahash` wants rustc 1.72 or later:
> 
> ```
> 11:54:16  error: package `ahash v0.8.8` cannot be built because it requires rustc 1.72.0 or newer, while the currently active rustc version is 1.67.0
> 11:54:16  Either upgrade to rustc 1.72.0 or newer, or use
> 11:54:16  cargo update -p ahash@0.8.8 --precise ver
> 11:54:16  where `ver` is the latest version of `ahash` supporting rustc 1.67.0
> ```

I'm slightly confused, because if you're using Bazel, then `ahash` should be pinned to v0.8.3 (even with this PR), and if you're not, then `rules_rust` shouldn't matter.

mikegrass:
> I'm slightly confused, because if you're using Bazel, then `ahash` should be pinned to v0.8.3 (even with this PR), and if you're not, then `rules_rust` shouldn't matter.

That's a good point. We are using bazel, and all the references in our repo and this one are to v0.8.3. I'm new to bazel; is there an easy way to query how a particular version of a dependency is being pulled in?

PiotrSikora:
> We are using bazel, and all the references in our repo and this one are to v0.8.3. I'm new to bazel; is there an easy way to query how a particular version of a dependency is being pulled in?

```
bazel query "deps(//:your_target)"
```

PiotrSikora:
But regardless of your issues with `ahash`, this PR is a good change to make (I already have this in my local tree, but didn't get around to pushing it yet).

mikegrass:
@PiotrSikora Hmm so it turns out our repo is using a bit of a frankenbuild, some parts building with bazel and the rust wasm filters not.

I'll look at cleaning that up so we don't hit surprise dependency upgrade issues like the error above in our build.

We are using this repo to generate portions of our build tooling though, so getting this PR merged will help us unblock while doing the larger build tooling cleanup.

Sounds like it's ready to merge?

antonengelhardt:
Just ran into the same issue and for me it helped to upgrade Rust to 1.75 in the Dockerfile (in case you are building with Docker)...

PiotrSikora:
> Sounds like it's ready to merge?

Almost. Could you also update `bazel-skylib`, since it's only used by `rules_rust`?

PiotrSikora:
> Done in the latest [commit](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/pull/221/commits/738a367caef497371561f393ef949d6d762d4821).

Thanks! Will get this in after #222.

PiotrSikora:
> Is that still the case? Do you happen to know of a way we could use Bazel builds while also stripping debuginfo?

It looks like you can strip it with `rustc_flags = ["-Cstrip=debuginfo"]`.

mikegrass:
@PiotrSikora I pulled in the changes from the now-merged #222. Can you please review?

PiotrSikora (approved):
Thanks!

PR #226, opened 2024-02-27
In-Code Documentation

antonengelhardt:
In-Code Documentation for the following functions:

* `dispatch_http_call()`
* `on_http_call_response()`
* `on_vm_start()`
* `get_vm_configuration()`
* `on_configure()`
* `get_plugin_configuration()`
* `set_tick_period()`
* `on_tick()`
* `on_http_request_headers()`
* `get_http_request_headers()`
* `get_http_request_header()`
* `add_http_request_header()`
* `resume_http_context()`
* `send_http_response()`


antonengelhardt:
All examples do not include the `RootContext` implementation or `main!` Macro, as they are pretty similar in every case. So copy and pasting them, won't work. I can still add them, if you think it's better.

antonengelhardt:
@PiotrSikora What do you think? Sorry to ping you, but I think this would be a great benefit for DX

PiotrSikora (commented):
> @PiotrSikora What do you think? Sorry to ping you, but I think this would be a great benefit for DX

I agree those would be very helpful. Sorry for the delay, I'll review this in the coming days.

src/traits.rs
```diff
@@ -523,6 +933,38 @@ pub trait HttpContext: Context {
...
+    /// # Example
+    ///
+    /// ```rust
+    /// use log::info;
+    /// use proxy_wasm::traits::*;
+    /// use proxy_wasm::types::*;
+    ///
+    /// impl HttpContext for MyHttpContext {
+    ///   fn on_http_request_headers(&mut self, _num_headers: usize, _end_of_stream: bool) -> Action {
+    ///     let auth = self.get_http_request_header("Authorization").unwrap_or_defauklt();
```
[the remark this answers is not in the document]
antonengelhardt: Sorry about that. I was a little careful when copy and pasting stuff, because it always messed up the doc formatting. But i have tested it ;)

src/traits.rs
```diff
@@ -523,6 +933,38 @@ pub trait HttpContext: Context {
...
+    ///       // Send an HTTP response with a status code of 200 and a body of "Hello, World!"
+    ///       self.send_http_response(200, vec![("A header", "Some Value")], Some(b"Hello, World!"));
+    ///     } else {
+    ///       // Send an HTTP response with a status code of 403 and a body of "Forbidden"
+    ///       self.send_http_response(307, vec![("location", "https://authenticate-here.com")], Some(b"Forbidden"));
+    ///     }
+    ///
+    ///     Action::Pause
+    ///   }
+    /// }
```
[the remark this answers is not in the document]
antonengelhardt: Fixed!

antonengelhardt:
> > @PiotrSikora What do you think? Sorry to ping you, but I think this would be a great benefit for DX
> 
> I agree those would be very helpful. Sorry for the delay, I'll review this in the coming days.

@PiotrSikora I tested all documented functions again and set up a [repo](https://github.com/antonengelhardt/proxy-wasm-playground) with 7 folders that cover all functions. It might still be that something is missing. I will do some diffs on the docs and the testing-repo-files to see if anything is missing apart from the boilerplate stuff (`proxy_wasm::main!`, RootContext Impl, etc.).

I hope this makes the review as easy as possible.

antonengelhardt:
I can squash my changes into one commit, if required.

antonengelhardt:
@PiotrSikora I can also add the examples from the mentioned repo to the `examples/` directory...

src/traits.rs
```diff
@@ -58,6 +58,67 @@ pub trait Context {
...
+    /// * `body` - The body of the request (in the form of `Option<&[u8]>` - `None` if no body)
+    /// * `trailers` - The trailers to send with the request (in the form of `Vec<(&str, &str)>`)
+    /// * `timeout` - The timeout for the request
+    ///
+    /// # Returns
+    ///
+    /// * `OK` on success.
+    /// * `BAD_ARGUMENT` for unknown upstream, or when headers are missing required `:authority`, `:method` and/or `:path` values.
+    /// * `INTERNAL_FAILURE' when the host failed to send requested HTTP call.
+    /// * `INVALID_MEMORY_ACCESS` when `upstream_data`, `upstream_size`, `headers_data`, `headers_size`, `body_data`, `body_size`, `trailers_data`, `trailers_size` and/or `return_call_id` point to invalid memory address.
```
PiotrSikora, line 78: It looks that you've copied those return values from the spec, which is generally fine... However, this SDK:
- combines `*_size` and `*_data` into vectors and strings,
- assumes that the memory objects are correct, so `INVALID_MEMORY_ACCESS` cannot happen and if it does, then it will result in `panic`,
- it's opinionated about which return values are allowed, so if you want to document them, then take a look at `src/hostcalls.rs` and see which return values are there.

src/traits.rs
```diff
@@ -58,6 +58,67 @@ pub trait Context {
...
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// use proxy_wasm::traits::*;
+    /// use proxy_wasm::types::*;
+    /// use std::time::Duration;
+    /// use log::warn;
```
PiotrSikora, line 86: I don't think that those examples are verified/build as part of the CI, since you're importing `warn` here, but using `info` in the code.

PiotrSikora:
> @PiotrSikora I can also add the examples from the mentioned repo to the `examples/` directory...

I took a look at them, and they seem to be mostly duplicates of the existing examples.

src/traits.rs
antonengelhardt: I found [this](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html). But i am not sure how i can test these examples without running envoy (in a container) and building/loading the examples into a wasm-plugin...

PiotrSikora: It's not about testing examples, but about making sure that the code in documentation even compiles.

PiotrSikora: @antonengelhardt it looks that you can use the `no_run` attribute to skip running of the code as part of the documentation testing:
```
/// ```no_run
/// ...code...
/// ```
```

antonengelhardt: Thanks for the hint! I am a bit busy with work and other things, but i will continue with these tests when i have time :)

antonengelhardt: @PiotrSikora The docs should compile now when running with `cargo test --doc`

src/traits.rs
antonengelhardt: I removed the docs from the spec and wrote something myself. Is that better?

.github/workflows/rust.yml
```diff
@@ -340,6 +340,29 @@ jobs:
...
+        rm rustup-init.sh
+        echo "$HOME/.cargo/bin" >> $GITHUB_PATH
+
+    - name: Update Rust
+      run: |
+        rustup toolchain install stable --component clippy --component rustfmt
+        rustup target add wasm32-wasi
+
+    - name: Run Cargo Doc Test
+      run: cargo test --doc
```
PiotrSikora, line 364: This should be added as an extra step into `msrv`, `stable` and `nightly` jobs.

src/traits.rs
PiotrSikora: I didn't mean that including text from the (now merged) spec is wrong. I meant that it had to be adjusted to account for the higher-level types used in the SDK.

I'll take a look at it in the next few days, thanks!

src/traits.rs
```diff
@@ -215,26 +330,245 @@ pub trait Context {
...
+    ///         serde_json::from_slice::<MyVmConfiguration>(&vm_configuration).unwrap();
+    ///
+    ///     // Do something with the parsed vm configuration
+    ///     debug!("vm_configuration: {:?}", parsed_vm_configuration);
+    ///
+    ///     true
+    ///   }
+    /// }
+    ///
+    /// # impl Context for MyRootContext {}
```
antonengelhardt, line 374: I was wondering if the `Context`-implementations, which are needed for the successful compilation, should be shown or hidden in the examples. @PiotrSikora

.github/workflows/rust.yml
antonengelhardt: Updated.

antonengelhardt:
@PiotrSikora Can you let the CI checks run so that i can see if it works?

PiotrSikora:
> @PiotrSikora Ok, so the Doc Test CI looks fine, but i dont know why the bazel job is failing. Thanks for running the CI ✌🏻

You've added new dependencies, so you need to tell Bazel about them. See [`DEVELOPMENT.md`](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/main/DEVELOPMENT.md).

antonengelhardt:
@PiotrSikora Can you please run them again?

antonengelhardt:
@PiotrSikora Friendly ping...

PiotrSikora:
@mpwarres @martijneken @leonm1 could you guys do a first pass? Thanks!

PR #229, opened 2024-03-25
Add metric related hostcalls to Context trait.

casimiro:
Also add an example filter demonstrating the use of metrics.


mvniekerk:
Hey all,
This could really be helpful what I'm busy with.

Anyone from Google looking at this?

PR #234, opened 2024-06-06
Rename the default branch to main.

martijneken:

PR #235, opened 2024-06-06
Fix build with Rust v1.80.0-nightly (2024-05-05).

martijneken:


.github/workflows/rust.yml
```diff
@@ -18,11 +18,11 @@ on:
...
-    - master
+    - main
 
   push:
     branches:
-    - master
+    - main
 
   schedule:
     - cron: '0 0 * * *'
```
PiotrSikora, line 28: Could you remove those changes, since they'll be merged in #234? Thanks!

PiotrSikora (dismissed):

.github/workflows/rust.yml
martijneken: Yep included these to get CI checks. Will remove when #234 is merged.

martijneken:
PTAL, https://github.com/proxy-wasm/proxy-wasm-rust-sdk/pull/236 fixes the last CI failure

PiotrSikora (approved):
Thanks!

PR #237, opened 2024-06-11
Accept more Status codes in get_property.

alexsnaps:
Adds the missing `Status` enum variants...
We currently run into `SerializationFailure` when getting some attribute `filter_state` from Envoy... Today, this not only panics, but the error isn't even really recognized. 
I'm not sure, given the current `trait Context` definition how to best deal with these gracefully tho? The default implementation are "just" `Result::unwrap()` ing the underlying host call.. So thinking that [this](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/pull/237/commits/64712b3d50dd0eb45db8de2af6be9e31bfb83ff2) is the best for now, wdyt?

### DIFF
--- a/src/hostcalls.rs
+++ b/src/hostcalls.rs
@@ -416,6 +416,8 @@ pub fn get_property(path: Vec<&str>) -> Result<Option<Bytes>, Status> {
                 }
             }
             Status::NotFound => Ok(None),
+            Status::SerializationFailure => Err(Status::SerializationFailure),
+            Status::InternalFailure => Err(Status::InternalFailure),
             status => panic!("unexpected status: {}", status as u32),
         }
     }
--- a/src/types.rs
+++ b/src/types.rs
@@ -46,14 +46,9 @@ pub enum Status {
     BadArgument = 2,
     SerializationFailure = 3,
     ParseFailure = 4,
-    BadExpression = 5,
-    InvalidMemoryAccess = 6,
     Empty = 7,
     CasMismatch = 8,
-    ResultMismatch = 9,
     InternalFailure = 10,
-    BrokenConnection = 11,
-    Unimplemented = 12,
 }
 
 #[repr(u32)]
--- a/src/types.rs
+++ b/src/types.rs
@@ -44,10 +44,16 @@ pub enum Status {
     Ok = 0,
     NotFound = 1,
     BadArgument = 2,
+    SerializationFailure = 3,
     ParseFailure = 4,
+    BadExpression = 5,
+    InvalidMemoryAccess = 6,
     Empty = 7,
     CasMismatch = 8,
+    ResultMismatch = 9,
     InternalFailure = 10,
+    BrokenConnection = 11,
+    Unimplemented = 12,
 }
 
 #[repr(u32)]


src/hostcalls.rs
```diff
@@ -416,7 +416,7 @@ pub fn get_property(path: Vec<&str>) -> Result<Option<Bytes>, Status> {
                 }
             }
             Status::NotFound => Ok(None),
-            status => panic!("unexpected status: {}", status as u32),
+            status => Err(status),
```
PiotrSikora, line 419: This catch-all panic for unexpected value is intentional in order to catch issues in host and/or SDK implementations, and it was invaluable in finding bugs in the past, so it should not be removed.

If there is a status value that is expected to work and it's missing, then it should be added explicitly.

PiotrSikora (changes requested):

src/hostcalls.rs
alexsnaps: _I think_ you are proposing to add explicit branches to the `match`, along side the existing `Ok` and `NotFound`, right? What would those be tho? I know `SerializationFailure` is definitely one of them, but I don't know what a host is allowed to return here? Is that spec'ed? 

But mostly confused about what you are trying to reach as a goal here? `panic` is a harsh fallback... Nothing the module can do about... Should we maybe bake the assumptions about what `Status` are expected in the default implementation for the `pub trait Context` ? So if anything unexpected happens, it still gets raised (as today it'd panic at `get_property(path).unwrap()`) while leaving a chance to the user to mitigate the issue?

"Happy path" remains `Ok` & `NotFound`, that seems obvious. `ParseFailure`, `InvalidMemoryAccess`, `CasMismatch`, `ResultMismatch` seem to be off the table at first glance... which leaves the others as possible outcomes?

src/hostcalls.rs
```diff
@@ -415,7 +417,11 @@ pub fn get_property(path: Vec<&str>) -> Result<Option<Bytes>, Status> {
                     Ok(None)
                 }
             }
-            Status::NotFound => Ok(None),
+            Status::NotFound | Status::Empty => Ok(None),
+            Status::SerializationFailure
+            | Status::BadArgument
+            | Status::BadExpression
+            | Status::Unimplemented => Err(status),
```
alexsnaps, line 424: This is a blind shot at trying to guess what "could" happen. I'm of the opinion that opening a seam to let the SDK user deal with them would be a more desirable way... if only "not the default" path so to keep the additional value of making it easier for implementers to test their host. 

Unsure what the best approach would be? The could also be some global "flag" or something to get to a more lenient runtime which lets the user take on the responsibility of dealing with "less standard" behavior maybe? 

If it is somewhere documented what valid `Status` every host method can return, I'm more than happy to generalize the pattern across the board. 

Or would alternate behaviors be encouraged to be supported by using a tweaked copy of this SDK instead?

src/hostcalls.rs
PiotrSikora: > _I think_ you are proposing to add explicit branches to the `match`, along side the existing `Ok` and `NotFound`, right? What would those be tho? I know `SerializationFailure` is definitely one of them, but I don't know what a host is allowed to return here? Is that spec'ed?

Looking at Envoy's code, the missing values are `InternalFailure` and `SerializationFailure`.

The return values are generally spec'ed out (see: https://github.com/proxy-wasm/spec/pull/42), but the properties in the current form are a bit hand-wavy, so we're using Envoy's code as the source of truth for v0.2.1.

> But mostly confused about what you are trying to reach as a goal here? `panic` is a harsh fallback... Nothing the module can do about...

The `panic` here is meant to handle the case when the host returned unexpected (invalid) status code.

It's not about handling the "sad path".

src/hostcalls.rs
PiotrSikora: > If it is somewhere documented what valid `Status` every host method can return, I'm more than happy to generalize the pattern across the board.

This pattern is already in this SDK, with some missing values (as you found out).

> Or would alternate behaviors be encouraged to be supported by using a tweaked copy of this SDK instead?

What alternative behaviors do you have in mind, exactly?

src/hostcalls.rs
alexsnaps: Changed as requested.

src/hostcalls.rs
alexsnaps: >What alternative behaviors do you have in mind, exactly?

A path where the client code could gracefully deal with the "out of spec" (based on the above, that'd be any host behaving differently than the current Envoy codebase tho). 
`panic!` in a crate is some radical thing to do. While I do see the advantage for this being the "default" behavior indeed, having some way to dealing gracefully with the `Err` would be nice I think... So unsure how to best enable that, wdyt? Would you agree that it'd be a nice to have? If so, how? make the `unsafe` method `pub` maybe and let someone deal with the entirety of the "problem"? Have some alternate method that gets you access to the raw `Result`? Possibly layering the code, where the "default public API" would call into the other `pub fn`, but deal explicitly with the `Err` behavior?

alexsnaps:
Is there some next release on the horizon that could contain this (and [the other changes](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/compare/96f791aaf86ef95326022280be987f19635eeaa5...89b48bd1c12a9330a07a5f0c963f9595fce93ef8#diff-1f3764730c411662c2396f030f8806cf3f27883695a0b67e689a8ea53b07a63b), possibly more in the pipe)?

PiotrSikora (changes requested):

src/hostcalls.rs
PiotrSikora: I understand what are you suggesting, but I'm still wondering who would ever use that and why... What host is  expected to be knowingly out of spec and for what purpose?

PiotrSikora:
> Is there some next release on the horizon that could contain this (and [the other changes](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/compare/96f791aaf86ef95326022280be987f19635eeaa5...89b48bd1c12a9330a07a5f0c963f9595fce93ef8#diff-1f3764730c411662c2396f030f8806cf3f27883695a0b67e689a8ea53b07a63b), possibly more in the pipe)?

Yes! This or more likely next week.

src/hostcalls.rs
```diff
@@ -414,8 +414,10 @@ pub fn get_property(path: Vec<&str>) -> Result<Option<Bytes>, Status> {
                 } else {
                     Ok(None)
                 }
-            }
+            },
```
PiotrSikora, line 417: Nit: this is unnecessary change and it doesn't match the existing style in the codebase.

src/hostcalls.rs
alexsnaps: We would use this, if only just on this very issue. Today we are stuck waiting on merge and release. So we need to fork in order to be able to make progress. I don't think (or say _hope_) any host would ever _knowingly_ go out of spec... but the spec is loosely defined and requires one to look at Envoy if I go by how this was addressed. Envoy 1.20 (i.e. the proxy, the host - and probably more versions) crashed on `SerializationFailure`... All that to say that mistakes happen and hopefully also then get fixed, but until that happens it is, if only for [us](https://kuadrant.io/) - that leverage this as part of bigger project then used by others - panicking is a very radical thing to do. So while I agree that hopefully we can find these issue easily while testing because of that very current behavior, being able to then also gracefully deal with them is desirable from a UX perspective.

src/hostcalls.rs
```diff
@@ -414,8 +415,9 @@ pub fn get_property(path: Vec<&str>) -> Result<Option<Bytes>, Status> {
                 } else {
                     Ok(None)
                 }
-            }
-            Status::NotFound => Ok(None),
```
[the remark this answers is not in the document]
alexsnaps: That's the kinda things I do when you come back from PTO 🤦

PiotrSikora (approved):
LGTM, thanks!

PR #101, opened 2021-04-13
Add support for gRPC streams.

Shikugawa:
Add gRPC stream support for Rust SDK. It includes any workaround around callout token table caused by #100 is not landed to master.

PR #240, opened 2024-07-01
Add support for nested gRPC callouts.

andytesti:
Panic with message "proxy-wasm-rust-sdk/src/dispatcher.rs:121:14:\nalready borrowed: BorrowMutError" is triggered when `dispatch_grpc_call()` is invoked from `on_grpc_call_response()`. The problem is originated by a [`self.grpc_callouts.borrow_mut()` being invoked inside an `if` condition](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/main/src/dispatcher.rs#L456), that keeps the `RefMut` alive until the `true` branch ends. 
This PR adopts the [same approach than on_http_call_response()](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/main/src/dispatcher.rs#L409), by invoking the `RefCell::borrow_mut()` outside the `if` condition.

### DIFF
--- a/src/dispatcher.rs
+++ b/src/dispatcher.rs
@@ -453,7 +453,8 @@ impl Dispatcher {
     }
 
     fn on_grpc_receive(&self, token_id: u32, response_size: usize) {
-        if let Some(context_id) = self.grpc_callouts.borrow_mut().remove(&token_id) {
+        let context_id = self.grpc_callouts.borrow_mut().remove(&token_id);
+        if let Some(context_id) = context_id {
             if let Some(http_stream) = self.http_streams.borrow_mut().get_mut(&context_id) {
                 self.active_id.set(context_id);
                 hostcalls::set_effective_context(context_id).unwrap();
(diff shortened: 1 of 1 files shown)


PiotrSikora (changes requested):
Thanks! This is a great find (and something that we've missed in #56).

However, this fix is incomplete, since it only covers `on_grpc_receive()` callback, and only for gRPC calls and not for gRPC streams.

Could you also fix the case for nested gRPC streams in `on_grpc_receive()` and for nested gRPC calls and streams in `on_grpc_close()`?

andytesti:
> Thanks! This is a great find (and something that we've missed in #56, EDIT: because it predates gRPC support added in #100 and #101).
> 
> However, this fix is incomplete, since it only covers `on_grpc_receive()` callback, and only for gRPC calls and not for gRPC streams.
> 
> Could you also fix the case for nested gRPC streams in `on_grpc_receive()` and for nested gRPC calls and streams in `on_grpc_close()`?

Thanks for reviewing this, @PiotrSikora. I just added fixes for nested gRPC streams and `on_grpc_close()` cases.

PiotrSikora (approved):
LGTM, thanks!

PR #242, opened 2024-07-11
Add simple stats API

junr03:
This PR exposes the concept of Metrics with a high level API. It takes liberal inspiration from the Simple<Counter,Gauge,Histogram> in the CPP SDK.

Users of the Rust SDK will now be able to create and manipulate simple Counters, Gauges, and Histograms that exist in the Envoy Runtime.

Note that this change is more generic than https://github.com/proxy-wasm/proxy-wasm-rust-sdk/pull/229 given that it exposes the concept of Counters, Gauges, and Histograms as a first class implementation, rather than interacting directly with the metrics host calls via the Root Context.

cc @casimiro

### DIFF
--- a/examples/metrics/README.md
+++ b/examples/metrics/README.md
@@ -1,0 +1,32 @@
+## Proxy-Wasm plugin example: HTTP headers
+
+Proxy-Wasm plugin that logs HTTP request/response headers.
+
+### Building
+
+```sh
+$ cargo build --target wasm32-wasi --release
+```
+
+### Using in Envoy
+
+This example can be run with [`docker compose`](https://docs.docker.com/compose/install/)
+and has a matching Envoy configuration.
+
+```sh
+$ docker compose up
+```
+
+Send HTTP request to `localhost:10000/`:
+
+```sh
+$ curl localhost:10000/ -H "x-envoy-wasm-metric-value: 100" -H "x-envoy-wasm-metric: gauge"
+```
+
+For instance that request will set the example gauge to 100. Which you can see using the stats endpoint
+
+```sh
+& curl -s localhost:9001/stats | grep wasmcustom.wasm_gauge
+
+100
+```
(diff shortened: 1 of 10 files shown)


src/stats.rs
```diff
@@ -0,0 +1,107 @@
...
+    fn id(&self) -> u32 {
+        self.id
+    }
+
+    fn name(&self) -> &str {
+        self.name.as_str()
+    }
+
+    fn record(&self, _: u64) {
+        // A Counter can only be incremented.
```
adilhafeez, line 44: Should we throw an exception here or maybe have separate trait called `MetricCounter` and place `increment` there.

adilhafeez (commented):
It would be great if you can update one of the examples and show use of stats there.

adilhafeez (approved):

examples/metrics/README.md
```diff
@@ -0,0 +1,32 @@
+## Proxy-Wasm plugin example: HTTP headers
```
Swaagie, line 1:
```suggestion
## Proxy-Wasm plugin example: Metrics
```

examples/metrics/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+Proxy-Wasm plugin that logs HTTP request/response headers.
```
Swaagie, line 3: Please update this as well.

examples/metrics/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Building
+
+```sh
+$ cargo build --target wasm32-wasi --release
```
Swaagie, line 8:
```suggestion
$ cargo build --target wasm32-wasip1 --release
```

Swaagie (changes requested):
I've actually tested this local on a project and it works nicely. @PiotrSikora what is your opinion on this vs the other metric PR?

PR #245, opened 2024-07-13
Add example using gRPC callouts.

PiotrSikora:


PiotrSikora:
A better version is included in #248.

PR #252, opened 2024-07-21
Update MSRV to Rust v1.64.0.

PiotrSikora:
Technically, this happened as a result of hashbrown update in #195,
but we didn't catch it at the time.

Add tests using MSRV to prevent this from happening in the future.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -88,6 +88,55 @@ jobs:
         bazelisk --noworkspace_rc run --noenable_bzlmod //bazel/cargo:crates_vendor
         git diff --exit-code
 
+  msrv:
+    runs-on: ubuntu-latest
+
+    env:
+      RUSTFLAGS: -D warnings
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install dependencies
+      if: ${{ env.ACT }}
+      run: |
+        curl -OL https://static.rust-lang.org/rustup/rustup-init.sh
+        chmod +x ./rustup-init.sh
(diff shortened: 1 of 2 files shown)


PiotrSikora:
Note that we published v0.2.2 with incorrect MSRV (1.61.0 instead of 1.64.0). Luckily, a higher MSRV (incorrectly declared as 1.63.0, and not 1.64.0) is declared in the `hashbrown` dependency, so this shouldn't be a big issue, so I won't be yanking our release unless there is a flood of user reports about it.

Also, v1.64.0 was released in September 2022, so hopefully people are using more recent versions by now...

.github/workflows/rust.yml
```diff
@@ -88,6 +88,55 @@ jobs:
...
+      run: cargo fmt -- --check
+
+    - name: Format (manifest)
+      run: cargo verify-project
+
+    - name: Package (docs)
+      run: cargo doc --no-deps --target=wasm32-unknown-unknown
+
+    - name: Package (publish)
+      run: cargo publish --dry-run --target=wasm32-unknown-unknown
```
mpwarres, line 138: Should --dry-run be removed?

mpwarres (approved):

.github/workflows/rust.yml
PiotrSikora: Nope (compare to `stable` and `nightly` jobs).

This test validates that the `cargo publish` would succeed, but it doesn't publish new crate on each run.

mpwarres: SG, thanks!

PR #256, opened 2024-08-28
Add convenience functions to remove headers and trailers.

itsLucario:
This is a convenient method to remove request header by key than using set_http_request_header with empty value. Since we already have the things in place, it's just about exposing it for usage with HttpContext.


src/hostcalls.rs
```diff
@@ -318,6 +318,15 @@ pub fn set_map_value(map_type: MapType, key: &str, value: Option<&str>) -> Resul
...
 
+pub fn remove_map_value(map_type: MapType, key: &str) -> Result<(), Status> {
+    unsafe {
+        match proxy_remove_header_map_value(map_type, key.as_ptr(), key.len()) {
+            Status::Ok => Ok(()),
+            status => panic!("unexpected status: {}", status as u32),
+        }
+    }
+}
+
```
PiotrSikora, line 329: Style: Could you move it above, so that it's not in-between `set_map_value` and `set_map_value_bytes`?

Ideally, could you group it with the `extern C` prototype, so that it's:
```
extern "C" {
    fn proxy_remove_header_map_value(
    [...]
}

pub fn remove_map_value(map_type: MapType, key: &str) -> Result<(), Status> {
[...]

extern "C" {
    fn proxy_replace_header_map_value(
    [...]
}

pub fn set_map_value(map_type: MapType, key: &str, value: Option<&str>) -> Result<(), Status> {
[...]
```

to match the existing style in this file?

src/traits.rs
```diff
@@ -351,6 +351,10 @@ pub trait HttpContext: Context {
         hostcalls::add_map_value_bytes(MapType::HttpRequestHeaders, name, value).unwrap()
     }
 
+    fn remove_http_request_header(&self, name: &str) {
```
PiotrSikora, line 354: Could you add those for request trailer, response header and response trailer as well? Thanks!

PiotrSikora (changes requested):

src/traits.rs
PiotrSikora: Friendly ping.

src/hostcalls.rs
PiotrSikora: Nit: The `remove` group should be before `set` group, since the latter is using `proxy_remove_header_map_value`, which is now used before declaration.

PiotrSikora (approved):
Thanks!

PR #265, opened 2024-10-11
Add Promise support for http callout

jizhuozhi:
In the current envoy WASM plugins, if we need IO request such as HTTP/GRPC/Redis, we must register request to envoy event loop and assigned a token via such as `dispatch_http_call` and WASM plugin should yield current request lifetime using `Action::Pause`. When the IO request completed, the envoy will callback to WASM plugin via `on_http_call_response`, and plugin should dispatch response using token (or ignored if single IO request). If we want to share something between `dispatch_http_call` and `on_http_call_response`, we must share them in plugin context fields, it's not a suitable scope.

Here is an example from [proxy-wasm-rust-sdk](https://github.com/proxy-wasm/proxy-wasm-rust-sdk/blob/main/examples/http_auth_random/src/lib.rs)
```rust
impl HttpContext for HttpAuthRandom {
    fn on_http_request_headers(&mut self, _: usize, _: bool) -> Action {
        self.dispatch_http_call(
            "httpbin",
            vec![
                (":method", "GET"),
                (":path", "/bytes/1"),
                (":authority", "httpbin.org"),
            ],
            None,
            vec![],
            Duration::from_secs(1),
        )
        .unwrap();
        Action::Pause
    }

    fn on_http_response_headers(&mut self, _: usize, _: bool) -> Action {
        self.set_http_response_header("Powered-By", Some("proxy-wasm"));
        Action::Continue
    }
}

impl Context for HttpAuthRandom {
    fn on_http_call_response(&mut self, _: u32, _: usize, body_size: usize, _: usize) {
        if let Some(body) = self.get_http_call_response_body(0, body_size) {
            if !body.is_empty() && body[0] % 2 == 0 {
                info!("Access granted.");
                self.resume_http_request();
                return;
            }
        }
        info!("Access forbidden.");
        self.send_http_response(
            403,
            vec![("Powered-By", "proxy-wasm")],
            Some(b"Access forbidden.\n"),
        );
    }
}
```

So there is three major problem we need to resolve:
1. How to easier dispatch request and callback response via token.
2. How to share something between dispatch request and callback response with smaller scope.
3. How to make code more fluid instead of spreading logic in different places

In Rust async programming, normally we use `async/await` for IO request, but in envoy WASM plugin, there is no executor to poll future. A suitable solution is providing JavaScript style Promise (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise). With Promise, we could write all logic in single function 
```Rust
impl HttpContext for HttpAuthRandom {
    fn on_http_request_headers(&mut self, _: usize, _: bool) -> Action {
        let token = self.dispatch_http_call(
            "httpbin",
            vec![
                (":method", "GET"),
                (":path", "/bytes/1"),
                (":authority", "httpbin.org"),
            ],
            None,
            vec![],
            Duration::from_secs(1),
        )
        .unwrap();
        let promise = Promise::new();
        // make relation between promise and request (token)
        promise.then(|(_, _, body_size, _)| {
            if let Some(body) = hostcalls::get_http_call_response_body(0, body_size) {
                if !body.is_empty() && body[0] % 2 == 0 {
                    info!("Access granted.");
                    hostcalls::resume_http_request();
                    return;
                }
            }
            info!("Access forbidden.");
            hostcalls::send_http_response(
                403,
                vec![("Powered-By", "proxy-wasm")],
                Some(b"Access forbidden.\n"),
            );
        })
        Action::Pause
    }

    fn on_http_response_headers(&mut self, _: usize, _: bool) -> Action {
        self.set_http_response_header("Powered-By", Some("proxy-wasm"));
        Action::Continue
    }
}
```
It seems more fluid then writing in callback of `on_http_call_response`, but there is no executor for promise to trigger state transferring. We can use `on_http_call_response` as trigger simply
```rust
struct HttpAuthRandom {
    promise: Rc<Promise<(u32, usize, usize, usize)>>
}

impl HttpContext for HttpAuthRandom {
    fn on_http_request_headers(&mut self, _: usize, _: bool) -> Action {
      // ...
      // make relation between promise and request (token)
      self.promise = promise.clone();
      // ...
    }
}

impl Context for HttpAuthRandom {
    fn on_http_call_response(&mut self, _token_id: u32, _num_headers: usize, _body_size: usize, _num_trailers: usize) {
        self.promise.fulfill((_token_id, _num_headers, _body_size, _num_trailers))
    }
}
```

As for making relationship between multi tokens and promises, we could just using `HashMap` with `insert/remove` (maybe it should be embed in SDK but not belongs to this PR)
```rust
struct HttpAuthRandom {
    m: HashMap<u32, Rc<Promise<(u32, usize, usize, usize)>>
}

impl HttpContext for HttpAuthRandom {
    fn on_http_request_headers(&mut self, _: usize, _: bool) -> Action {
      // ...
      // make relation between promise and request (token)
      self.m.insert(token, promise.clone());
      // ...
    }
}

impl Context for HttpAuthRandom {
    fn on_http_call_response(&mut self, _token_id: u32, _num_headers: usize, _body_size: usize, _num_trailers: usize) {
        let promise = self.m.remove(token);
        promise.fulfill((_token_id, _num_headers, _body_size, _num_trailers))
    }
}
```


examples/http_parallel_call/src/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+        ])
+        .then(|results| {
+            send_http_response(
+                200,
+                vec![],
+                Some(
+                    format!(
+                        "{}{}\n",
+                        results[0].strip_suffix("\n").unwrap(),
+                        results[1].strip_suffix("\n").unwrap()
```
PiotrSikora, line 88: Nit: You should not request `\n` from `httpbin.org` if you're going to strip it here. Adds unnecessary noise.

examples/http_parallel_call/src/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+        self.m.insert(token1, promise1.clone());
+        self.m.insert(token2, promise2.clone());
+
+        Promise::all_of(vec![
+            promise1
+                .then(|(_, _, _body_size, _)| get_http_call_response_body_string(0, _body_size))
+                .then(|body| body.unwrap_or_else(|| "".to_string())),
+            promise2
+                .then(|(_, _, _body_size, _)| get_http_call_response_body_string(0, _body_size))
+                .then(|body| body.unwrap_or_else(|| "".to_string())),
```
PiotrSikora: Style: `_` prefix is used for variables that are not used. Since you're using `_body_size`, you shouldn't prefix it with `_`.

examples/http_parallel_call/src/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+                None,
+                vec![],
+                Duration::from_secs(1),
+            )
+            .unwrap();
+
+        let promise1 = Promise::new();
+        let promise2 = Promise::new();
+        self.m.insert(token1, promise1.clone());
+        self.m.insert(token2, promise2.clone());
```
PiotrSikora, line 70: This is leaking a lot of async machinery into the plugin's codebase, which means that it becomes public API that we need to maintain, and that users need to handle it themselves.

Instead, we should be handling this within the SDK.

examples/http_parallel_call/src/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+        &mut self,
+        _token_id: u32,
+        _num_headers: usize,
+        _body_size: usize,
+        _num_trailers: usize,
+    ) {
+        let promise = self.m.remove(&_token_id);
+        promise
+            .unwrap()
+            .fulfill((_token_id, _num_headers, _body_size, _num_trailers));
```
PiotrSikora, line 115: Same style comment about the use of `_` prefix.

jizhuozhi:
`msrv/stable/nightly` has been fixed, but another `outdated/reactors` seems not belong to this PR, PTAL :)

src/promise.rs
```diff
@@ -0,0 +1,338 @@
...
+                });
+        }
+        next_promise
+    }
+
+    pub fn any_of(promises: Vec<Rc<Self>>) -> Rc<Promise<T>> {
+        let next_promise = Promise::new();
+        let total = promises.len();
+        let remaining = Rc::new(RefCell::new(total));
+        let first_error = Rc::new(RefCell::new(None)); // 用来保存第一个错误
```
PiotrSikora, line 149: Please stick to English, thanks!

src/promise.rs
```diff
@@ -0,0 +1,338 @@
...
+                assert!(results.is_empty());
+                *touched_clone.borrow_mut() = true;
+            })
+            .catch(|_err| {
+                panic!("Should not reach here");
+            });
+
+        assert_eq!(true, touched.take())
+    }
+}
```
PiotrSikora, line 341: This makes me a bit nervous. I don't think this SDK should be in the business of maintaining a generic async machinery like that... However, I've never liked the existing flow, and I've been looking into adding something like this myself over the last week (for sync/async KV store), so this is definitely a welcomed feature. Thanks!

I'm dealing with something else right now, but I'll try to review it later this week.

PiotrSikora:
> `msrv/stable/nightly` will fix, but another `outdated/reactors` seems not belong to this PR, PTAL :)

Thanks! I'll fix existing errors shortly. Sorry about that!

examples/http_parallel_call/src/lib.rs
jizhuozhi: I have implemented a dispatcher to wrap `dispatch_http_call` in my project, but not submitted in proxy-wasm. Does any suggested style or file organized in proxy-wasm? Otherwise I will take it to `dispatcher.rs` to handle this within the SDK :)

```rust
#[derive(Default)]
pub struct HttpDispatcher {
    m: HashMap<u32, Rc<Promise<(u32, usize, usize, usize)>>>,
}

impl HttpDispatcher {
    pub fn dispatch(
        &mut self,
        upstream: &str,
        headers: Vec<(&str, &str)>,
        body: Option<&[u8]>,
        trailers: Vec<(&str, &str)>,
        timeout: Duration,
    ) -> Rc<Promise<(u32, usize, usize, usize)>> {
        let token =
            hostcalls::dispatch_http_call(upstream, headers, body, trailers, timeout).unwrap();
        let promise = Promise::new();
        self.m.insert(token, promise.clone());
        promise
    }

    pub fn callback(
        &mut self,
        _token_id: u32,
        _num_headers: usize,
        _body_size: usize,
        _num_trailers: usize,
    ) {
        let promise = self.m.remove(&_token_id);
        promise
            .unwrap()
            .fulfill((_token_id, _num_headers, _body_size, _num_trailers))
    }
}
```

src/promise.rs
jizhuozhi: Sorry, my mistake, I will remove all Chinese comments

jizhuozhi:
There is currently a problem involving the collaboration between SDK and user code: `HttpClient` is created by user code, but the `callback` is driven by the SDK. There is currently no way to shield this detail within the SDK. User code always needs to write boilerplate code in `on_http_call_response` to driver `callback`.

I have no idea to resolve this problem :(

jizhuozhi:
Baze BUILD file has been reformatted :)

PiotrSikora:
> There is currently a problem involving the collaboration between SDK and user code: `HttpClient` is created by user code, but the `callback` is driven by the SDK. There is currently no way to shield this detail within the SDK. User code always needs to write boilerplate code in `on_http_call_response` to driver `callback`.

I think (but didn't check the code) that this boilerplate could be included in the dispatcher, and executed automatically when events for registered callouts are received from the host... or am I missing something?

src/callout/promise.rs
```diff
@@ -0,0 +1,341 @@
+// Copyright 2024 Google LLC
```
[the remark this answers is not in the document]
jizhuozhi: `std::future` in c++ is an async primitive to get result in multi-thread programming. `std::future` produced by `std::async` or `std::promise`. Once program calls `std::future::get`, it will block until `std::async` completed or `std::promise::set` called. It means at least two threads for provider and consumer but wasm has no (once call dispatched, thread will be used to process other requests).
And `future-rs` is providing the foundations for asynchronous programming in Rust, and lazy executed which requires an executor to poll the result (which means all operations must be non-blocking). The problem as same as`std::future`, there is no thread as executor to poll the result.
So the only way that we could own the thread is as callback of http response.

jizhuozhi:
> > There is currently a problem involving the collaboration between SDK and user code: `HttpClient` is created by user code, but the `callback` is driven by the SDK. There is currently no way to shield this detail within the SDK. User code always needs to write boilerplate code in `on_http_call_response` to driver `callback`.
> 
> I think (but didn't check the code) that this boilerplate could be included in the dispatcher, and executed automatically when events for registered callouts are received from the host... or am I missing something?

The Go SDK handles it this way, but I'm not sure if doing so breaks the Rust SDK API

PR #271, opened 2024-10-18
robustness

pi-pi-miao:


src/hostcalls.rs
```diff
@@ -25,7 +26,10 @@ pub fn log(level: LogLevel, message: &str) -> Result<(), Status> {
     unsafe {
         match proxy_log(level, message.as_ptr(), message.len()) {
             Status::Ok => Ok(()),
-            status => panic!("unexpected status: {}", status as u32),
+            status => {
+                error!("[log] unexpected status: {}", status as u32);
+                Err(status)
+            },
```
PiotrSikora, line 32: This is done by design, since those errors are illegal are not actionable. See #282.

src/hostcalls.rs
```diff
@@ -1199,48 +1356,189 @@ mod utils {
...
+                String::new() // 或者根据你的需求返回默认值
+            });
+
+            let value_string = String::from_utf8(value).unwrap_or_else(|e| {
+                error!("Failed to convert value to String: {}", e);
+                String::new() // 或者根据你的需求返回默认值
+            });
+
+            // 将成功转换的键值对推入 map
+            map.push((key_string, value_string));
```
PiotrSikora, line 1446: Map is serialized by a trusted host environment, so it's not untrusted/malicious input. See: #282.

PiotrSikora (commented):
Hi @pi-pi-miao, the CI should be fixed now. Sorry for the delay!

As for your changes, may I ask what motivated them?

You should never see those unknown statuses from the Proxy-Wasm compliant proxies. See: #282.

As for UTF-8 `String` failures, those are indeed a problem, but using either `_bytes()` variants (which were introduced as a solution to that) or a `HeaderValue` type (#287) is a better solution than returning errors and/or ignoring non-UTF-8 values.

I believe all other changes are a viral effect of propagating errors from the 2 issues above, but please correct me if I've missed something.

Thanks!

PiotrSikora:
Closing per comments. Also, see: #287, #282, #319.

PR #273, opened 2024-10-23
Update hashbrown to v0.15.

PiotrSikora:


bazel/cargo/Cargo.Bazel.lock
```diff
@@ -42,77 +37,10 @@ version = "0.4.22"
...
-
 [[package]]
 name = "proxy-wasm"
 version = "0.2.2"
 dependencies = [
  "hashbrown",
  "log",
 ]
-
-[[package]]
```
mpwarres, line 68: I take it these are just automated updates due to changes in hashbrown dependencies?

mpwarres (approved):

bazel/cargo/Cargo.Bazel.lock
PiotrSikora: Correct, and the CI enforces that as well.

PR #276, opened 2024-10-23
Fix GitHub Actions for examples when running locally.

PiotrSikora:


PiotrSikora:
Actually, let's skip those tests when testing locally. See #280.

PR #279, opened 2024-10-25
Add instructions for running GitHub Actions locally.

PiotrSikora:
While there, rename example and reactor jobs to their plural forms.


README.md
```diff
@@ -28,10 +28,29 @@
 - [Extending Envoy with WASM and Rust](https://antweiss.com/blog/extending-envoy-with-wasm-and-rust/)
```
PiotrSikora, line 28: I'm not sure if this should be in the README.md or a separate file linked from here, since it doesn't apply to consumers of the SDK?

mpwarres: Either way SGTM. If you put it in a separate file, DEVELOPMENT.md would make sense. We don't currently have that in the C++ SDK but could update to match.

mpwarres (approved):

PR #284, opened 2024-10-28
Pass slice instead of Vec during map serialization.

PiotrSikora:
Before:

utils::bench_serialize_map         ... bench:    66.87 ns/iter (+/- 1.33)
utils::bench_serialize_map_bytes   ... bench:    69.42 ns/iter (+/- 1.45)

After:

utils::bench_serialize_map         ... bench:    55.84 ns/iter (+/- 1.16)
utils::bench_serialize_map_bytes   ... bench:    56.03 ns/iter (+/- 1.19)

Note that this is changed only in the private functions to avoid breaking
the public API.

### DIFF
--- a/src/hostcalls.rs
+++ b/src/hostcalls.rs
@@ -190,7 +190,7 @@ extern "C" {
 }
 
 pub fn set_map(map_type: MapType, map: Vec<(&str, &str)>) -> Result<(), Status> {
-    let serialized_map = utils::serialize_map(map);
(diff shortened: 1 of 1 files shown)


PiotrSikora:
This is stacked on #283, since it requires updates in tests and benchmarks, so please look only at the 2nd commit in this PR.

mpwarres (approved):

Swaagie (approved):

PR #286, opened 2024-10-28
Stop using "our" Bytes type.

PiotrSikora:
It's a simple alias to Vec<u8>, so it doesn't provide any value,
and it conflicts with a popular Bytes type from the bytes crate.

This is a non-breaking change, since we retain the public alias.


mpwarres (approved):

Swaagie (approved):

leonm1 (approved):

leonm1:
Good to merge from a Google perspective (I don't have write permissions for the rust-sdk, otherwise I'd merge it myself.)

PR #287, opened 2024-10-28
Use RFC-compliant HeaderValue type.

PiotrSikora:
When "header-value" feature is enabled, a RFC-compliant HeaderValue
type is used instead of the previously used UTF-8 String type.

Since HTTP field values aren't UTF-8 encoded, using the String type
meant that a plugin would crash when valid, but obsolete, non-UTF-8
characters were present in HTTP headers and/or trailers.

This feature is currently optional to avoid breaking changes and to
help with migration, but it will become a default feature in v0.3.

The HeaderValue type is re-exported from the http crate.


Cargo.toml
```diff
@@ -11,9 +11,15 @@ edition = "2018"
...
 
 [dependencies]
+bytes = { version = "1", optional = true }
 hashbrown = "0.15"
+http = { version = "1", optional = true }
 log = "0.4"
 
+[features]
+default = []
+header-value = ["dep:bytes", "dep:http"]
```
PiotrSikora, line 21: This feature could be renamed to `valid-header`, `strict-header` or `rfc-header` to make this reusable in case we want to add restrictions on header name in the future.

Alternatively, we could drop this as a feature and instead add those functions as variants with `_typed()` suffix.

Thoughts?

src/hostcalls.rs
```diff
@@ -1159,46 +1196,60 @@ mod utils {
...
-        let mut bytes: Bytes = Vec::with_capacity(size);
-        bytes.extend_from_slice(&map.len().to_le_bytes());
-        for (name, value) in &map {
-            bytes.extend_from_slice(&name.len().to_le_bytes());
-            bytes.extend_from_slice(&value.len().to_le_bytes());
+        serialize_map(map)
+    }
+
+    #[cfg(feature = "header-value")]
+    pub(super) fn deserialize_map(mut bytes: bytes::Bytes) -> Vec<(String, HeaderValue)> {
```
PiotrSikora, line 1227: Note to self: drop `bytes::` prefix if merged after #286.

Cargo.toml
mpwarres: `_typed()` suffix would fit alongside the existing `_bytes()` suffix, so doesn't seem too bad to me.

No objection to keeping it as a feature, though, assuming we find a solution to the HTTP/2 pseudo-headers problem. If you keep it a feature, `strict-header` or `rfc-header` SGTM.

PiotrSikora:
@mpwarres could we get someone from Google to convert Rust tests in Envoy to use the `header-value` feature (it's already exposed via `@proxy_wasm_rust_sdk//:proxy_wasm_header_value` Bazel target), to verify that I didn't miss anything obvious, before we merge this PR? I converted some tests when I originally wrote this PR, which was a while ago, so the basics do work, but it wasn't a rewrite of everything.

mpwarres:
> @mpwarres could we get someone from Google to convert Rust tests in Envoy to use the `header-value` feature (it's already exposed via `@proxy_wasm_rust_sdk//:proxy_wasm_header_value` Bazel target), to verify that I didn't miss anything obvious, before we merge this PR? I converted some tests when I originally wrote this PR, which was a while ago, so the basics do work, but it wasn't a rewrite of everything.

Sorry, I somehow missed this earlier. Yes, we will work on this now.

PiotrSikora:
> Sorry, I somehow missed this earlier. Yes, we will work on this now.

Thanks!

Cargo.toml
PiotrSikora: Done.

src/hostcalls.rs
```diff
@@ -219,6 +237,38 @@ extern "C" {
...
+        match proxy_get_header_map_value(
+            map_type,
+            key.as_ptr(),
+            key.len(),
+            &mut return_data,
+            &mut return_size,
+        ) {
+            Status::Ok => {
+                if !return_data.is_null() {
+                    match HeaderValue::from_bytes(std::slice::from_raw_parts(
```
PiotrSikora, line 254: This would leak memory (see: #324).

src/hostcalls.rs
```diff
@@ -145,6 +145,27 @@ extern "C" {
...
+#[cfg(feature = "strict-header-value")]
+pub fn get_map(map_type: MapType) -> Result<Vec<(String, HeaderValue)>, Status> {
+    unsafe {
+        let mut return_data: *mut u8 = null_mut();
+        let mut return_size: usize = 0;
+        match proxy_get_header_map_pairs(map_type, &mut return_data, &mut return_size) {
+            Status::Ok => {
+                if !return_data.is_null() {
+                    let serialized_map =
+                        bytes::Bytes::from(std::slice::from_raw_parts(return_data, return_size));
```
PiotrSikora, line 157: This would leak memory (see: #324).

PiotrSikora:
@mpwarres @leonm1 PTAL, thanks!

src/hostcalls.rs
```diff
@@ -191,6 +192,28 @@ pub fn get_map_bytes(map_type: MapType) -> Result<Vec<(String, Bytes)>, Status>
...
+                        return_data,
+                        return_size,
+                        return_size,
+                    ));
+                    Ok(utils::deserialize_map_typed(serialized_map))
+                } else {
+                    Ok(Vec::new())
+                }
+            }
+            status => panic!("unexpected status: {}", status as u32),
```
leonm1, line 219: Side note: don't forget to use maybe_panic in this PR if merged after no-panics

src/hostcalls.rs
```diff
@@ -288,6 +332,35 @@ pub fn get_map_value_bytes(map_type: MapType, key: &str) -> Result<Option<Bytes>
...
+        match proxy_get_header_map_value(
+            map_type,
+            key.as_ptr(),
+            key.len(),
+            &mut return_data,
+            &mut return_size,
+        ) {
+            Status::Ok => {
+                if !return_data.is_null() {
+                    // We're intentionally using the unchecked variant in order to retain
```
leonm1, line 357: The http crate docs indicate a HeaderValue containing non-UTF-8 characters is undefined behavior: https://docs.rs/http/latest/http/header/struct.HeaderValue.html#method.from_maybe_shared_unchecked

I suppose this is fine if to_str returns an error instead of panicing, but this feels just as risky if users start to use HeaderValues assuming they're valid UTF-8.

Why not bubble up the error on line 295 when creating the string from the value returned from the host?

PiotrSikora: > The http crate docs indicate a HeaderValue containing non-UTF-8 characters is undefined behavior: https://docs.rs/http/latest/http/header/struct.HeaderValue.html#method.from_maybe_shared_unchecked

Thanks! That's a good catch.

However, that comment is almost definitely wrong - the code path in debug builds checks for valid HTTP field value characters, not UTF-8 characters, and the whole reason for `HeaderValue` type is that the value is not UTF-8 in the first place.

I'm less clear on what the author had in mind by the "undefined behavior" comment, since I don't see anything that would result in such thing, even with invalid header field values stored inside the `HeaderValue`, since `to_str()` returns `&str` only if it the whole value is a printable subset of the allowed characters, and returns error otherwise.

Also, it looks that Cloudflare's Pingora is [accepting invalid header chars](https://github.com/cloudflare/pingora/blob/faac65b0c2a0bfdbfdc5f13a1591f53f3c15321a/pingora-core/src/protocols/http/v1/server.rs#L1297-L1300) and adding them into [`HeaderValue::from_maybe_shared_unchecked`](https://github.com/cloudflare/pingora/blob/faac65b0c2a0bfdbfdc5f13a1591f53f3c15321a/pingora-core/src/protocols/http/v1/server.rs#L248-L251) in builds with `patched_http1` feature. See their [test cases](https://github.com/cloudflare/pingora/blob/faac65b0c2a0bfdbfdc5f13a1591f53f3c15321a/pingora-core/src/protocols/http/v1/server.rs#L1376-L1387).

> I suppose this is fine if to_str returns an error instead of panicing, but this feels just as risky if users start to use HeaderValues assuming they're valid UTF-8.

But they cannot misuse `HeaderValue` as a valid UTF-8 `String`, since `to_str()` returns value only for it's a printable subset of valid characters (which are also valid UTF-8 characters), and errors out otherwise.

> Why not bubble up the error on line 295 when creating the string from the value returned from the host?

The whole reason for this PR and existence of `HeaderValue` type is that HTTP field values can contain characters that are not valid UTF-8, so we panic on `String::from_utf8()` that contains perfectly legitimate HTTP field value.

Basically, Rust's UTF-8 `String` type is a wrong type for the job.

The reasons why I oped into using `HeaderValue::from_maybe_shared_unchecked` (i.e. without enforcing strict RFC compliance) here is that browsers and proxies are generally very relaxed when it comes to accepting invalid HTTP header values, with both Chrome and NGINX accepting everything else than `0x00` character, and we don't want to be in situation in which adding plugin to the request path breaks the traffic.

Using `HeaderValue::from_maybe_shared_unchecked` means that we retain those non-compliant values, while we're not producing new ones. Note that those values were already accepted by the host's parsers.

See [RFC9110 spec for HTTP field values](https://httpwg.org/specs/rfc9110.html#fields.values):
```text
field-value    = *field-content
field-content  = field-vchar
                 [ 1*( SP / HTAB / field-vchar ) field-vchar ]
field-vchar    = VCHAR / obs-text
obs-text       = %x80-FF
```
 
See [WHATWG spec for HTTP header values](https://fetch.spec.whatwg.org/#terminology-headers):
```
A header value is a byte sequence that matches the following conditions:
- Has no leading or trailing HTTP tab or space bytes.
- Contains no 0x00 (NUL) or HTTP newline bytes.
```

which also warns that:
```
The definition of -WHATWG- header value is not defined in terms of the -IETF- field-value token production as it is not compatible with deployed content.
```

PiotrSikora: Filed https://github.com/hyperium/http/pull/822, while also asking to define "undefined behaviour".

Also, this function is not `unsafe` in Rust sense, see: https://github.com/hyperium/http/pull/102#issue-248481396.

leonm1: Thanks for filing an upstream issue! I'm good to merge this as-is given the discussion there.

leonm1 (approved):

PR #297, opened 2024-12-31
Adjust GitHub Actions for Ubuntu 24.04.

PiotrSikora:
GitHub migrated `ubuntu-latest` to `ubuntu-24.04`, which comes with
a different set of pre-installed software. Most notably, both cargo
audit and outdated tools are missing.

Pin jobs to a specific version of the runner image to avoid another
upgrade from breaking us in the future.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -30,7 +30,7 @@ on:
 jobs:
 
   licenses:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
 
     steps:
     - uses: actions/checkout@v2
@@ -53,7 +53,7 @@ jobs:
         addlicense -check -ignore "bazel/cargo/remote/**" .
 
   bazel:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
 
     steps:
     - uses: actions/checkout@v2
@@ -98,7 +98,7 @@ jobs:
         git diff --exit-code
 
   msrv:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
 
     env:
       RUSTFLAGS: -D warnings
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -505,15 +494,6 @@ jobs:
     - name: Format (manifest)
       run: cargo verify-project
 
-    - name: Run cargo audit
```
leonm1, line 508: It looks like this PR removes auditing for both examples + reactors (though keeps the SDK itself `cargo audit`ed). Is that intended? If so, could you leave a comment as to express this happened and indicate that it is unnecessary?

PiotrSikora: It's the same reasoning as to why they were excluded in local tests in #280 (i.e. the cost of building those tools doesn't provide too much value in examples, so it's not worth building them if they're not pre-installed as in full `ubuntu-22.04` images on GitHub Actions).

With that in mind, I moved them to the `audit` and `outdated` jobs, since we build those tools for the SDK already.

mpwarres (approved):

.github/workflows/rust.yml
```diff
@@ -46,14 +46,17 @@ jobs:
           ~/go/pkg/mod
         key: licenses-${{ hashFiles('.github/workflows/rust.yml') }}
 
-    - name: Check licenses
+    - name: Install dependencies
       run: |
         go install github.com/google/addlicense@latest
```
PiotrSikora, line 51: One thing that's bothering me is that the CI has been breaking lately due to updates to tools and build image.

I pinned the build image to `ubuntu-24.04` in this PR, but should we also pin all the tools (`addlicense`, `bazelisk`, `buildifier` `cargo-audit`, cargo-outdated`), and possibly even Rust toolchains for `stable` and `nightly` jobs (which would require more frequent PRs with version bumps)?

PR #299, opened 2024-12-31
Fix conditional attributes in exported allocator name.

PiotrSikora:
`#[export_name]` already takes precedence over `#[no_mangle]`,
so this is a no-op, but it silences the warning.

Found with nightly.


leonm1 (approved):
Good fix, thanks for sending a PR!

PR #305, opened 2025-02-28
Differentiate between missing and empty headers.

prembhaskal:
The method `get_http_request_headers_bytes` return empty vector for empty header value but `get_http_request_header_bytes` returns None for empty header value.   

Similarly `get_http_request_headers` returns empty String but `get_http_request_header` returns None.

This commit intends to fix this inconsistency.

With this, it will be possible to differentiate between two cases, a missing header and a header with empty value.

### DIFF
--- a/src/hostcalls.rs
+++ b/src/hostcalls.rs
@@ -241,7 +241,7 @@ pub fn get_map_value(map_type: MapType, key: &str) -> Result<Option<String>, Sta
                         .unwrap(),
                     ))
                 } else {
-                    Ok(None)
+                    Ok(Some(String::new()))
                 }
             }
             Status::NotFound => Ok(None),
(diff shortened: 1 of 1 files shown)


PiotrSikora:
Which host returns `NULL` value here?

This code is a legacy code path for "header not found" from Proxy-Wasm ABI v0.1.0, and I'd expect that hosts would return a valid pointer to a zero-sized string, although I guess returning `NULL` here is also fine.

Also, you need to make the same change in `get_map_value_bytes`.

(Please ignore the CI failures, they'll get fixed once #306 is merged)

prembhaskal:
@PiotrSikora I think `pub fn get_map_value` is the legacy one, which returns utf8 string value. I have not changed that, but please let me know if I should.

prembhaskal:
@PiotrSikora I have updated the method `get_map_value` now.

prembhaskal:
Hi @PiotrSikora can you please rerun test, I see the PR to fix workflow is merged.

PiotrSikora (approved):
Thanks! And sorry for the delay.

PiotrSikora:
@prembhaskal could you merge with the `main` branch? It looks that I don't have permission to do that myself.

prembhaskal:
Thanks @PiotrSikora I rebased with `main` branch now.

PR #324, opened 2025-07-29
Fix memory leak in get_map and get_map_bytes functions.

JaatadiaMulesoft:


JaatadiaMulesoft:
I was testing compiling the filter and deploying it, after some request you can clearly see that the memory starts to grow. Now checking If I can add a unit test, I've found several libraries that could be used for this [mockalloc](https://docs.rs/mockalloc/latest/mockalloc/) and [allocation_counter](https://docs.rs/allocation-counter/latest/allocation_counter/), etc. Any suggestion?

JaatadiaMulesoft:
Test added

PiotrSikora:
> I was testing compiling the filter and deploying it, after some request you can clearly see that the memory starts to grow. Now checking If I can add a unit test, I've found several libraries that could be used for this [mockalloc](https://docs.rs/mockalloc/latest/mockalloc/) and [allocation_counter](https://docs.rs/allocation-counter/latest/allocation_counter/), etc. Any suggestion?

Apologies for the delay.

The `mockalloc` test works fine for this particular case, but I was looking for a more generic solution for finding memory leaks (e.g. running sanitizers) using existing tests and/or examples.

Notably, #287 also incorrectly uses `std::slice::from_raw_parts()`, but one leak wouldn't be found with changes in this PR, and generally speaking, I'm not a big fan of using mocks instead of integration tests, so I think it would be better if we could run this against a real host implementation... and if we want to use mocks anyway, then it should cover more than a single function.

Having said that, that's a much bigger undertaking, and it shouldn't block this fix, so could you revert this PR to a version without the `mockalloc`? Sorry for the noise, I should have merged it right away without blocking it on tests.

joesbigidea:
Hi! I notice this has been waiting for a while and I'm interested in this fix as well. I don't want to steal the thunder for this fix but would be happy to open a new PR with just the fix and without the additional test if that helps get this through.

PiotrSikora (approved):
Thanks!

mpwarres (approved):

Issue #310, opened 2025-03-27
Panics in the SDK


PiotrSikora:
@alexsnaps said in https://github.com/proxy-wasm/proxy-wasm-rust-sdk/pull/282#issuecomment-2748139891:

> This issue with these `panic` is that currently envoy doesn't play nicely with them... Tho [things](https://github.com/envoyproxy/envoy/pull/36456) are getting better, it can still result in unexpected behavior. So while I understand the rational behind wanting to `panic!`, given this is wasm and panic are aborting execution, this leaves the caller with no way to gracefully handle the error. Which by itself is problematic in my perspective, but together with some environment then failing further (like starving workers), is really not acceptable, as I imagine we can all agree on.
> 
> So, should there be a way to provide the user with means to be "fully responsible"? it could be "as simple" as allowing the user to call into functions that would expose all failures, while leaving the trait et al default implementation panic? Or maybe add a "SDK ~panic~ error handler" that defaults to the current behavior, but would let the user override that behavior? Thinking out loud here, but while I understand ideally "it would all always work according to spec", its hard to guarantee it will, more so with an on-going specification.
> 
> This has come up in the past when discussing one of my PRs with @PiotrSikora . We are now running into issues in production that don't allow us to tolerate the SDK panicking. Looking into a viable path forward that would check everyone's goals, if such a thing is possible obviously. Thanks!

PiotrSikora:
> This issue with these `panic` is that currently envoy doesn't play nicely with them... Tho [things](https://github.com/envoyproxy/envoy/pull/36456) are getting better, it can still result in unexpected behavior. So while I understand the rational behind wanting to `panic!`, given this is wasm and panic are aborting execution, this leaves the caller with no way to gracefully handle the error. Which by itself is problematic in my perspective, but together with some environment then failing further (like starving workers), is really not acceptable, as I imagine we can all agree on.

`panics` are only used in the code paths that cannot happen (unless there is a bug in the SDK or Envoy, but those should be fixed in those projects, and not in the plugins).

However, it sounds like you're saying that those `panics` are a regular occurrences, but I don't see too many reports about them from you (other than #237 that got fixed a while ago), so I'm a little confused about the frequency and how it's really affecting you in production.

Also, how do you plan to handle those cases that should never happen gracefully in your plugin?

[...]

PR #282, opened 2024-10-28
Document the trust model and panics in the codebase.

PiotrSikora:

PR #345, opened 2025-12-14
Remove unsafe unwraps in hostcalls.rs

foo4u:
Fix unsafe string data handling when converting from the proxy ABI to Rust. This change replaces `String::from_utf8(..).unwrap()` calls with `String::from_utf8_lossy(..)`, which does not panic and adds zero runtime cost when strings are valid UTF-8. Furthermore, ABI to Rust code has been de-duplicated and made to return an error instead of panicking if conversion fails.

Future API changes could be made to not unwap this error, providing further panic protection if desired.

This change resolves the user-facing issues being discussed in #310, Panics in the SDK.


src/hostcalls.rs
```diff
@@ -156,32 +156,26 @@ fn proxy_get_header_map_pairs(
...
-                    let serialized_map = Vec::from_raw_parts(return_data, return_size, return_size);
-                    Ok(utils::deserialize_map(&serialized_map))
-                } else {
-                    Ok(Vec::new())
-                }
-            }
-            status => panic!("unexpected status: {}", status as u32),
-        }
-    }
+    get_map_impl(map_type, |v| String::from_utf8_lossy(&v).into_owned())
```
PiotrSikora, line 159: `String::from_utf8_lossy()` is a wrong "solution" - it changes `panic` to a silently corrupted data (which I could argue is even worse), and it doesn't actually solve the underlying problem (UTF-8 strings are wrong the type to use for HTTP field values).

If you want to use `String::from_utf8_lossy()`, then `_bytes()` variants provide an escape hatch and allow converting opaque bytes to other types inside plugins and not in the SDK. Those variants have been available for years.

However, the proper solution is to use the appropriate type for HTTP field values, see: #287 (reviews welcome!).

src/hostcalls.rs
```diff
@@ -1265,49 +1275,64 @@ mod utils {
...
-            ));
-        }
-        map
-    }
 
-    pub(super) fn deserialize_map_bytes(bytes: &[u8]) -> Vec<(String, Bytes)> {
-        if bytes.is_empty() {
-            return Vec::new();
+        if bytes.len() < 4 {
+            return Err(Error::BufferTooShort);
```
PiotrSikora, line 1291: This generally makes sense for deserialization of untrusted data in networked protocols, but we're doing function calls within the same process, see: #282 (but also #319).

PiotrSikora (commented):
Hi @foo4u, sorry for the late review.

PiotrSikora:
Closing per comments. Also, see: #287, #282, #319.

src/hostcalls.rs
foo4u: @PiotrSikora I adamantly disagree that crashing the WASM runtime is better than parsing lossy. I wouldn't use a web server that crashed on bad inputs. Most web servers will return 4xx errors on invalid headers, but this project is for software embedded inside a proxy like Envoy, you cannot make assumptions. I actually only took this approach because you seemed reluctant to other input on changing function contracts.

The premise that panicking is acceptable on invalid user input opens software reliant on this library up to denial of service attacks. Quite frankly I was shocked to see panics on invalid user input in the library. You mention I could use `_bytes()` but based on other discussions here I see others assuming the library's string parsing is safe when it's anything but. We don't all read library source code before using it.

I will look at your solution in #287, but panicking on invalid user input is never acceptable in a library crate. I really think you should be open to changing that ASAP.

PS - my issue is not about this PR, but about the cavalier attitude by maintainer(s) ignoring Rust best practices. Libraries should never panic based on invalid user inputs. See https://github.com/proxy-wasm/proxy-wasm-rust-sdk/issues/310.

src/hostcalls.rs
foo4u: I understand the point in #282 but it doesn't negate that you should validate your inputs.

src/hostcalls.rs
PiotrSikora: @foo4u I'm not sure where you got the idea that I think that anything (servers, libraries, whatever) should crash on invalid user input. That's definitely not the case.

foo4u: >  it changes panic to a silently corrupted data (which I could argue is even worse), 

That's sounds like panicking is better than accepting malformed data. The correct approach on these functions would be to return `Result<String, Error>`.

PR #362, opened 2026-03-07
Use rust_shared_library for building Wasm modules with Bazel.

PiotrSikora:
The rust_binary(out_binary=True) hack is no longer needed,
and we can remove the patch for it.


leonm1 (approved):
Great news, thanks Piotr!

PR #366, opened 2026-05-19
Update Rust edition to 2024.

PiotrSikora:
This raises MSRV to Rust v1.85.0.

PR #378, opened 2026-06-23
Use protox to compile protobufs in examples.

PiotrSikora:
This avoids the dependency on the protobuf compiler.


PiotrSikora:
Failed `msrv` check will be addressed with MSRV update in #366.

PiotrSikora:
cc @SiiiTschiii

leonm1 (approved):

